Issue #11, opened 2021-04-26
Handle sources.yaml file


flavio:
Right now the user has to provide it via the cli flag. We should default it to the user `~/.kubewarden` directory.

The user should still be able to provide a different one via the already existing flag.

Issue #1, opened 2021-04-22
kwctl: Initial implementation


ereslibre:
Create the `kwctl` tool. This CLI tool is the go-to tool for Kubewarden users. It wil allow to perform:

- Policy image handling
  - Pull (from OCI registry, HTTP server...) -- akin to `docker pull`
  - Pull and run -- akin to `docker run`
  - Push/Publish (to OCI registry) -- akin to `docker push`
    - Being capability focused, the publishing mechanism will take the Wasm artifact, and a specification of what resources the policy can handle. This information is added to the final artifact, and finally published.
 - Generate Kubernetes CRD scaffolding
   - Given a Wasm artifact with permissions, the Kubernetes CRD scaffolding will generate a `ClusterAdmissionPolicy` that is ready to be applied to a Kubernetes cluster, assuming it has Kubewarden installed.

This is a mid-time overview, but the immediate action is to split the `policy-server` project into different crates and executables:

- crate: [`policy-evaluator`](https://github.com/kubewarden/policy-server/tree/main/crates/policy-evaluator)
- crate: [`policy-fetcher`](https://github.com/kubewarden/policy-server/tree/main/crates/policy-fetcher)
- executable: [`policy-testdrive`](https://github.com/kubewarden/policy-server/tree/main/crates/policy-testdrive) -> `kwctl` (this tool)

ereslibre:
The e2e tests and unit tests are going to be added right after. Pull requests open:

- policy-evaluator (optional), version bumps: https://github.com/kubewarden/policy-evaluator/pull/2
- policy-fetcher: https://github.com/kubewarden/policy-fetcher/pull/2
- kwctl (depends on the previous ones): https://github.com/kubewarden/kwctl/pull/13
- policy-server: https://github.com/kubewarden/policy-server/pull/58

PR #13, opened 2021-04-30
feature: implement basic feature set and CLI interface

ereslibre:
Fixes: https://github.com/kubewarden/kwctl/issues/1
Fixes: https://github.com/kubewarden/kwctl/issues/11

The interface looks as follows:

```
USAGE:
    kwctl <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    help        Prints this message or the help of the given subcommand(s)
    policies    Lists all downloaded policies
    pull        Pulls a Kubewarden policy from a given URI
    run         Runs a Kubewarden policy from a given URI
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,8 +13,9 @@ edition = "2018"
 [dependencies]
 anyhow = "1.0"
 clap = "2.33.3"
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", branch = "settings-options" }
-policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.1.0" }
+directories = "3.0.2"
+policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.1.1" }
+policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.1.3" }
 serde_json = "1.0"
 serde_yaml = "0.8.17"
 serde = { version = "1.0", features = ["derive"] }
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,6 @@
 extern crate anyhow;
+extern crate clap;
+extern crate directories;
 extern crate policy_evaluator;
 extern crate policy_fetcher;
 extern crate serde_yaml;
@@ -7,10 +9,16 @@ use anyhow::{anyhow, Result};
 use clap::{
     clap_app, crate_authors, crate_description, crate_name, crate_version, AppSettings, ArgMatches,
 };
-use std::{fs, path::PathBuf, str::FromStr};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+    str::FromStr,
+};
 
 use policy_fetcher::registry::config::{DockerConfig, DockerConfigRaw};
 use policy_fetcher::sources::{read_sources_file, Sources};
+use policy_fetcher::store::DEFAULT_ROOT;
+use policy_fetcher::PullDestination;
 
 mod policies;
 mod pull;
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -13,9 +13,10 @@ edition = "2018"
 [dependencies]
 anyhow = "1.0"
 clap = "2.33.3"
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", branch = "settings-options" }
-policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.1.0" }
+policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", branch = "update-dependencies" }
+policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", branch = "refactor" }
```
ereslibre, line 17: This is not intended to be merged, this PR depends on the following PR's:

- https://github.com/kubewarden/policy-evaluator/pull/2
- https://github.com/kubewarden/policy-fetcher/pull/2

src/main.rs
```diff
@@ -28,26 +31,26 @@ async fn main() -> Result<()> {
             )
             (@subcommand pull =>
              (about: "Pulls a Kubewarden policy from a given URI")
-             (@arg ("docker-config-json-path"): --("docker-config-json-path") +takes_value "Path to a Docker config.json-like path. Can be used to indicate registry authentication details")
-             (@arg ("sources-path"): --("sources-path") +takes_value "YAML file holding source information (https, registry insecure hosts, custom CA's...)")
+             (@arg ("docker-config-json-path"): default_value("~/.kubewarden/config.json") --("docker-config-json-path") +takes_value "Path to a Docker config.json-like path. Can be used to indicate registry authentication details")
+             (@arg ("sources-path"): default_value("~/.kubewarden/sources.yaml") --("sources-path") +takes_value "YAML file holding source information (https, registry insecure hosts, custom CA's...)")
```
flavio, line 35: The default values are going to be invalid on Windows and OSX

src/main.rs
```diff
@@ -28,26 +31,26 @@ async fn main() -> Result<()> {
...
-             (@arg ("uri"): * +takes_value "Policy URI. Supported schemes: registry://, https://, file://")
+             (@arg ("uri"): * "Policy URI. Supported schemes: registry://, https://, file://")
             )
             (@subcommand run =>
              (about: "Runs a Kubewarden policy from a given URI")
-             (@arg ("docker-config-json-path"): --("docker-config-json-path") +takes_value "Path to a Docker config.json-like path. Can be used to indicate registry authentication details")
-             (@arg ("sources-path"): --("sources-path") +takes_value "YAML file holding source information (https, registry insecure hosts, custom CA's...)")
+             (@arg ("docker-config-json-path"): default_value("~/.kubewarden/config.json") --("docker-config-json-path") +takes_value "Path to a Docker config.json-like path. Can be used to indicate registry authentication details")
+             (@arg ("sources-path"): default_value("~/.kubewarden/sources.yaml") --("sources-path") +takes_value "YAML file holding source information (https, registry insecure hosts, custom CA's...)")
```
flavio, line 42: The default values are going to be invalid on Windows and OSX

flavio (changes requested):
I don't like the usage of the `~/.kubewarden` directory. Even by using the `shellexpand` crate I think we end up having something that is doing unexpected things on platforms like Windows and OSx. Also on Linux, I would argue that this approach goes against the XDG specification.

I did some quick searches and I found the [directories](https://crates.io/crates/directories) crate which seems to be doing what we need, in a cross platform way.

PR #30, opened 2021-05-25
feat: allow to skip scheme on `kwctl run` command

ereslibre:
If `kwctl run` command is called with a module without a scheme,
assume the `file://` scheme, rooted on the local directory. This way,
it's possible to run modules in the following ways:

- `kwctl run registry://ghcr.io/kubewarden/policies/safe-labels:v0.1.2 -s <settings> -r <request>`
- `kwctl run file://$(pwd)/safe-labels.wasm -s <settings> -r <request>`
- `kwctl run safe-labels.wasm -s <settings> -r <request>`

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,7 +1,27 @@
 use anyhow::{anyhow, Result};
 use policy_fetcher::store::Store;
-use std::path::PathBuf;
+use regex::Regex;
+use std::{env, path::PathBuf};
 use url::Url;
+
+pub(crate) fn map_path_to_uri(uri: &str) -> Result<String> {
+    let uri_has_schema = Regex::new(r"^\w+://").unwrap();
+    if uri_has_schema.is_match(uri) {
+        return Ok(String::from(uri));
+    }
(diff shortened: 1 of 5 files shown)


src/utils.rs
```diff
@@ -17,3 +37,62 @@ pub(crate) fn wasm_path(uri: &str) -> Result<PathBuf> {
...
+            map_path_to_uri("some-policy-0.0.1.wasm")?,
+            format!(
+                "file://{}",
+                env::current_dir()?
+                    .join("some-policy-0.0.1.wasm")
+                    .into_os_string()
+                    .into_string()
+                    .map_err(|_| anyhow!("cannot get policy test path"))?,
+            ),
+        );
```
flavio, line 79: This is duplicate of the previous lines

flavio (changes requested):
LGTM, I just left a minor comment

src/utils.rs
ereslibre: Ooof, you are absolutely right

ereslibre:
I'll merge on green.

PR #31, opened 2021-05-27
Introduce "pretty" inspect output

flavio:
The `kwctl inspect` command now produces colorized and formatted output.

It's still possible to produce simple YAML outpout by using the `-o yaml` flag.

**Note well:** this branch depends on a currently unreleased version of the `policy-evaluator` crate, see https://github.com/kubewarden/policy-evaluator/pull/7

## Input metadata

The next examples take a policy that was annotated using this metadata:

```yaml
rules:
- apiGroups: [""]
  apiVersions: ["v1"]
  resources: ["pods"]
  operations: ["CREATE", "UPDATE"]
mutating: false
labels:
  production: false
annotations:
  name.castelli.hello: world
  io.kubewarden.policy.title: psp-apparmor
  io.kubewarden.policy.description: Replacement for the Kubernetes Pod Security Policy that controls the usage of AppArmor profiles
  io.kubewarden.policy.author: Flavio Castelli
  io.kubewarden.policy.url: https://github.com/kubewarden/psp-apparmor
  io.kubewarden.policy.source: https://github.com/kubewarden/psp-apparmor
  io.kubewarden.policy.license: Apache-2.0
  io.kubewarden.policy.usage: |
    This policy works by defining a whitelist of allowed AppArmor profiles. Pods are then inspected at creation and update time, to ensure only approved profiles are used.

    When no AppArmor profile is defined, Kubernetes will leave the final choice to the underlying container runtime. This will result in using the default AppArmor profile provided by Container Runtime. Because of that, the default behaviour of this policy is to accept workloads that do not have an AppArmor profile specified.

    The policy can be configured with the following data structure:
    ```yaml
    # list of allowed profiles
    allowed_profiles:
    - runtime/default
    - localhost/my-special-workload
    ```
```

## YAML output

The `kwctl inspect --output yaml ...` command produces the old yaml output:

![image](https://user-images.githubusercontent.com/22728/119826958-e3538300-bef8-11eb-9368-f43ecb42250c.png)

## New "pretty" output

By default this is the output produced by `kwctl inspect`:

![image](https://user-images.githubusercontent.com/22728/119827046-febe8e00-bef8-11eb-9cbd-627ca6f2b21c.png)

### DIFF
--- a/src/inspect.rs
+++ b/src/inspect.rs
@@ -1,16 +1,16 @@
-use std::convert::TryFrom;
-
 use anyhow::{anyhow, Result};
 use kubewarden_policy_sdk::metadata::ProtocolVersion;
 use mdcat::{ResourceAccess, TerminalCapabilities, TerminalSize};
 use policy_evaluator::policy_metadata::Metadata;
 use prettytable::{format::FormatBuilder, Table};
 use pulldown_cmark::{Options, Parser};
+use std::convert::TryFrom;
 use syntect::parsing::SyntaxSet;
 
-pub(crate) fn inspect(uri: &str, output: Option<&str>) -> Result<()> {
-    let wasm_path = crate::utils::wasm_path(uri)?;
-    let printer = get_printer(OutputType::try_from(output)?);
+pub(crate) fn inspect(uri: &str, output: OutputType) -> Result<()> {
+    let uri = crate::utils::map_path_to_uri(uri)?;
+    let wasm_path = crate::utils::wasm_path(uri.as_str())?;
+    let printer = get_printer(output);
 
     match Metadata::from_path(&wasm_path)? {
         Some(metadata) => printer.print(&metadata),
@@ -21,7 +21,7 @@ pub(crate) fn inspect(uri: &str, output: Option<&str>) -> Result<()> {
     }
 }
 
-enum OutputType {
+pub(crate) enum OutputType {
     Yaml,
     Pretty,
 }
@@ -139,10 +139,9 @@ impl MetadataPrettyPrinter {
     fn print_metadata_usage(&self, metadata: &Metadata) -> Result<()> {
         let usage = match metadata.annotations.clone() {
             None => None,
-            Some(annotations) => match annotations.get("io.kubewarden.policy.usage") {
-                Some(usage) => Some(String::from(usage)),
-                None => None,
-            },
+            Some(annotations) => annotations
+                .get("io.kubewarden.policy.usage")
+                .map(String::from),
         };
 
         if usage.is_none() {
--- a/src/inspect.rs
+++ b/src/inspect.rs
@@ -1,18 +1,197 @@
-use anyhow::{anyhow, Result};
-use policy_evaluator::policy_metadata::Metadata;
+use std::convert::TryFrom;
 
-pub(crate) fn inspect(uri: &str) -> Result<()> {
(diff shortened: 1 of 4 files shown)


ereslibre:
This PR needs a rebase :(

src/main.rs
```diff
@@ -83,6 +83,7 @@ async fn main() -> Result<()> {
             (@subcommand inspect =>
              (about: "Inspect Kubewarden policy")
              (@arg ("uri"): * "Policy URI. Supported schemes: registry://, https://, file://")
+             (@arg ("output"): -o --("output") +takes_value "output format. One of: yaml")
```
ereslibre, line 86: Hm, seems that the `clap_app!` macro does not have a way to provide a set of valid values. I can check later if there's a nice way we can do this.

src/inspect.rs
```diff
@@ -1,18 +1,161 @@
...
+    let parser = Parser::new_ext(
+        text,
+        Options::ENABLE_TASKLISTS | Options::ENABLE_STRIKETHROUGH,
+    );
+    let env = mdcat::Environment::for_local_directory(&std::env::current_dir()?)?;
+
+    let stdout = std::io::stdout();
+    let mut output = stdout.lock();
+    mdcat::push_tty(&settings, &env, &mut output, parser).or_else(|error| {
+        if error.kind() == std::io::ErrorKind::BrokenPipe {
```
ereslibre, line 155: How can this happen?

src/inspect.rs
```diff
@@ -1,18 +1,161 @@
 use anyhow::{anyhow, Result};
+use kubewarden_policy_sdk::metadata::ProtocolVersion;
+use mdcat::{ResourceAccess, TerminalCapabilities, TerminalSize};
 use policy_evaluator::policy_metadata::Metadata;
+use prettytable::{format::FormatBuilder, Table};
+use pulldown_cmark::{Options, Parser};
+use syntect::parsing::SyntaxSet;
 
-pub(crate) fn inspect(uri: &str) -> Result<()> {
+pub(crate) fn inspect(uri: &str, output: Option<&str>) -> Result<()> {
```
ereslibre, line 9: I would be in favour of having `output: OutputType`, being `OutputType` an enum that has a `TryFrom<&str>` and enumeration values of `Pretty` and `Yaml` for example. Thus, on the CLI side, we can construct a OutputType::try_from(output_type)`, being `output_type` the argument from the user. This way, this call can focus on the enum values we allow instead of having to match on `Option<&str>`.

ereslibre (approved):
LGTM, some bike shedding that would improve the interface a little. I think we can do more ergonomic things where it makes sense :)

src/inspect.rs
flavio: I have to say, this is coming from the mdcat source code. I think it could happen if `kwctl inspect` is piped into somethink like `less`

src/inspect.rs
flavio: Good idea! I'll work on that

flavio:
@ereslibre I pushed some changes. I think I went a bit extreme with the solution :sweat_smile:

src/inspect.rs
ereslibre: This `output` can now take an `OutputType`, right? :)

ereslibre (approved):
LGTM, thanks :)

PR #36, opened 2021-05-29
Handle the annotatedBy field of policy metadata

flavio:
Ensure kwctl writes its version when annotating a policy.

At the same time, ensure `inspect`ion of a policy shows who annotated it.

### DIFF
--- a/src/annotate.rs
+++ b/src/annotate.rs
@@ -1,4 +1,5 @@
 use anyhow::{anyhow, Result};
+use kubewarden_policy_sdk::metadata::ProtocolVersion;
 use policy_evaluator::{
     constants::KUBEWARDEN_CUSTOM_SECTION_METADATA, policy_evaluator::PolicyEvaluator,
     policy_metadata::Metadata,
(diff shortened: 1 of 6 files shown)


src/annotate.rs
```diff
@@ -27,6 +27,7 @@ fn prepare_metadata(wasm_path: PathBuf, metadata_path: PathBuf) -> Result<Metada
         .map_err(|e| anyhow!("Cannot compute ProtocolVersion used by the policy: {:?}", e))?;
 
     metadata.protocol_version = Some(protocol_version);
+    metadata.annotated_by = Some(format!("kwctl v{}", env!("CARGO_PKG_VERSION")));
```
ereslibre, line 31: If we need to process this on a later operation I think it better to write `CARGO_PKG_VERSION`, so we can just use a semver parser directly and do comparisons.

ereslibre (changes requested):

src/annotate.rs
flavio: resolved as discussed on slack (see below)

flavio:
I've done the following change to the PR:  kwctl creates an annotation inside of the policy's annotations to store its version.

The annotation is named `io.kubewarde.kwctl` and the version of kwctl is stored as value.

src/inspect.rs
```diff
@@ -78,12 +78,12 @@ impl MetadataPrettyPrinter {
...
-            "io.kubewarden.policy.author",
-            "io.kubewarden.policy.url",
-            "io.kubewarden.policy.source",
-            "io.kubewarden.policy.license",
+            crate::constants::ANNOTATION_POLICY_TITLE,
+            crate::constants::ANNOTATION_POLICY_DESCRIPTION,
+            crate::constants::ANNOTATION_POLICY_AUTHOR,
+            crate::constants::ANNOTATION_POLICY_URL,
+            crate::constants::ANNOTATION_POLICY_SOURCE,
+            crate::constants::ANNOTATION_POLICY_LICENSE,
```
ereslibre, line 86: By `use crate::constants` we could remove some of the duplication here and use `constants::THE_CONST`

ereslibre (approved):
LGTM, non-blocking bike shedding

src/inspect.rs
flavio: fixed

PR #38, opened 2021-05-31
Ensure the docker configuration is read as well

flavio:
Previous to this commit, kwctl relied only on the contents of `~/.config/kubewarden/config.json`. The contents of `~/.docker/config.json` were ignored.

This patch gives high priority to kubewarden's configuration file but, when this doesn't exists, attempts to fall back on docker's
configuration file.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -13,6 +13,7 @@ use anyhow::{anyhow, Result};
 use clap::{
     clap_app, crate_authors, crate_description, crate_name, crate_version, AppSettings, ArgMatches,
 };
+use directories::UserDirs;
 use std::{
     convert::TryFrom,
     fs,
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -244,12 +248,22 @@ fn remote_server_options(matches: &ArgMatches) -> Result<(Option<Sources>, Optio
...
+            let mut configuration_files = vec![DEFAULT_ROOT.config_dir().join("config.json")];
+            if let Some(user_dir) = UserDirs::new() {
+                let config_json_path = user_dir.home_dir().join(".docker").join("config.json");
+                configuration_files.push(config_json_path);
+            }
+
+            let mut conf: Option<DockerConfig> = None;
+            for path in configuration_files.iter() {
+                if Path::exists(&path) {
+                    conf = Some(read_docker_config_json_file(&path)?);
```
ereslibre, line 262: What happens if one configuration is not compatible, but a subsequent one is? We would bubble up the exception on the first one, right?

ereslibre (changes requested):

PR #39, opened 2021-06-03
Policy testdrive missing features

flavio:
This PR brings back two features that got lost during the `policy-testdrive` -> `kwctl` rewrite. They are both related with the behaviour of the `run` command:

1. Allow user to specify the policy settings on the cli, via the new `--settings-json` flag. This makes the process of writing bat tests easier
2. Ensure the settings provided by the user are actually validated. This must happen **before** the request is evaluated. Invalid settings lead `kwctl` to exit with an error


flavio:
@ereslibre once this is merged I'll tag a new release of kwctl

ereslibre (approved):
LGTM, thanks!

Issue #41, opened 2021-06-09
Extend the `push` command


flavio:
The `push` command should allow to push a pre-pulled policy to a different registry. This allows users to push an official policy to a local registry they own.

These are the proposed changes.

>`kwctl push --policy-path <file> <registry url>`

This pushes a local wasm file to a remote registry. There's no need to add the `file://` prefix. The path to the file can also be relative to the directory from which the command is invoked.

Example
`kwctl push --policy-path foo.wasm registry.com/foo:latest`

>`kwctl push --policy <uri> <registry url>`

This can be used to push a policy that has been previously pulled on the local machine.

The `uri` parameter requires the protocol to be specified. The protocol can be `file://`, `registry://`, `http://` and `http://`.
The push command will exit with an error if the specified policy hasn't been previously pulled.

Examples
  * `kwctl push --policy file:///var/lib/policies/foo.wasm registry.com/foo:latest`
  * `kwctl push --policy registry://ghcr.io/kubewarden/foo:latest registry.com/foo:latest`

PR #42, opened 2021-06-09
Extend the `push` command

flavio:
Change the push command to take two positional parameters:

  1. The policy to be pushed (source)
  2. The registry destination (destination)

This allows to push not only local files, but also policies that are
inside of the local storage.

This makes possible to push an upstream policy into a local registry.

Fixes https://github.com/kubewarden/kwctl/issues/41

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,6 +21,7 @@ use std::{
     str::FromStr,
 };
 
+use tracing::debug;
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::{fmt, EnvFilter};
 
@@ -62,10 +63,10 @@ async fn main() -> Result<()> {
             (@subcommand push =>
              (about: "Pushes a Kubewarden policy to an OCI registry")
              (@arg ("docker-config-json-path"): --("docker-config-json-path") +takes_value "Path to a Docker config.json-like path. Can be used to indicate registry authentication details")
-             (@arg ("policy-path"): * -p --("policy-path") +takes_value "Policy file to push")
-             (@arg ("sources-path"): --("sources-path") +takes_value "YAML file holding source information (https, registry insecure hosts, custom CA's...)")
-             (@arg ("uri"): * "Policy URI. Supported schemes: registry://")
              (@arg ("force"): -f --("force") "push also a policy that is not annotated")
+             (@arg ("sources-path"): --("sources-path") +takes_value "YAML file holding source information (https, registry insecure hosts, custom CA's...)")
+             (@arg ("policy"): * "Policy to push. Can be the path to a local file, or a policy URI")
+             (@arg ("uri"): * "Policy URI. Supported schemes: registry://")
             )
             (@subcommand rm =>
              (about: "Removes a Kubewarden policy from the store")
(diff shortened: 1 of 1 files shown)


flavio:
The e2e tests are broken for other reasons, I'm going to fix them ASAP

src/main.rs
```diff
@@ -62,8 +62,8 @@ async fn main() -> Result<()> {
             (@subcommand push =>
              (about: "Pushes a Kubewarden policy to an OCI registry")
              (@arg ("docker-config-json-path"): --("docker-config-json-path") +takes_value "Path to a Docker config.json-like path. Can be used to indicate registry authentication details")
-             (@arg ("policy-path"): * -p --("policy-path") +takes_value "Policy file to push")
              (@arg ("sources-path"): --("sources-path") +takes_value "YAML file holding source information (https, registry insecure hosts, custom CA's...)")
+             (@arg ("policy"): * "Policy to push. Can be the path to a local file, or a policy URI")
              (@arg ("uri"): * "Policy URI. Supported schemes: registry://")
              (@arg ("force"): -f --("force") "push also a policy that is not annotated")
```
ereslibre, line 68: I wonder if we should keep named arguments alphabetically sorted. `--force` was already off though.

src/main.rs
```diff
@@ -136,11 +136,24 @@ async fn main() -> Result<()> {
...
+                    .map(|u| {
+                        if u.starts_with("registry://") {
+                            String::from(u)
+                        } else {
+                            format!("registry://{}", u)
+                        }
+                    })
+                    .unwrap();
+
+                println!("Attempting to push {:?} to {:?}", wasm_path, uri);
```
ereslibre, line 152: I'd be silent by default, and maybe use some logging library with verbosity, structured logging... I think we can start building on that story.

src/main.rs
```diff
@@ -149,7 +162,7 @@ async fn main() -> Result<()> {
                     }
                 }
 
-                push::push(&policy, uri, docker_config, sources).await?;
+                push::push(&policy, uri.as_str(), docker_config, sources).await?;
```
ereslibre, line 165:
```suggestion
                push::push(&policy, &uri, docker_config, sources).await?;
```

It's personal taste again, I favour `&uri`, but whatever you prefer :)

ereslibre (changes requested):
Thanks, looks great. Some comments.

src/main.rs
```diff
@@ -62,8 +62,8 @@ async fn main() -> Result<()> {
             (@subcommand push =>
              (about: "Pushes a Kubewarden policy to an OCI registry")
              (@arg ("docker-config-json-path"): --("docker-config-json-path") +takes_value "Path to a Docker config.json-like path. Can be used to indicate registry authentication details")
-             (@arg ("policy-path"): * -p --("policy-path") +takes_value "Policy file to push")
              (@arg ("sources-path"): --("sources-path") +takes_value "YAML file holding source information (https, registry insecure hosts, custom CA's...)")
+             (@arg ("policy"): * "Policy to push. Can be the path to a local file, or a policy URI")
```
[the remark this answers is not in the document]
flavio: This is now a positional argument, like `uri`. According to my tests everything works fine

src/main.rs
flavio: I can sort the flag alphabetically

src/main.rs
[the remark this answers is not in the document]
ereslibre: Right, perfect!

flavio:
@ereslibre: I've submitted some fixes. Once you're fine with the PR I'll squash and merge

ereslibre (approved):

Issue #43, opened 2021-06-09
Allow to provide the request contents on `stdin`


ereslibre:
By using `kwctl --request-path -` it's possible to pipe a request in the `kwctl` stdin file descriptor. This allows us to do runs like:

`curl https://some-sample-request.json | kwctl --settings-json '' --request-path - registry://the-registry.com/the-policy:latest`

PR #45, opened 2021-06-09
feat: provide request on stdin

ereslibre:
By calling `kwctl run` with `-r -`, the request can be piped into
kwctl for ease of use.

Fixes https://github.com/kubewarden/kwctl/issues/43


flavio (approved):
LGTM

Issue #40, opened 2021-06-07
Fix cluster awareness logic


ereslibre:
Cluster awareness needs to be fixed, so that:

- When a policy that is not cluster aware is executed, `kwctl` does not try to initialize a Kubernetes client.
- When a policy that is cluster context aware is executed, `kwctl` does the following:
  - Fetch the cluster information
  - Execute policy once the cluster information has been retrieved
  - Exit

Despite `kwctl` and `policy-server` share some cluster aware code, they behave in a different way:

- `kwctl` by default does not fetch the cluster context, and thus, does not initialize a Kubernetes client, unless the metadata of the policy signals that the policy is cluster aware.
- `kwctl` does not initialize a context refresh loop.
- If the policy is context aware, `kwctl` initializes the cluster context with an initial request, and then runs the policy in the Wasm environment. waPC endpoints on the host expose the information to the guest, because the policy is not started until the context is fetchec from the cluster.

The `policy-server` initializes the cluster context, and keeps an infinite loop, refreshing the context information on a shared global variable, protected by `RwLocks`. This ensures that there's only one writer at any given time, and N readers.

ereslibre:
I cannot reproduce the `segfault`. The previous approach makes `kwctl` print an error on `stderr` but

flavio:
Sounds good to me. Are you going to handle the changes to policy metadata?

ereslibre:
If we are good with the patches and everything gets merged:

- I will release a new tag of `policy-evaluator` and bump the dependency on `kwctl` and `policy-server`. Release both.
- I will update all policies to include the new metadata, for now all setting `clusterAware: false`.

flavio:
Sounds like a good plan.

BTW, publishing a new policy-server container image will put us on the latest build of `opensuse/leap:15.3`, which right now has 0 vulnerabilities :partying_face:

PR #49, opened 2021-06-15
feat: initialize cluster context only if the policy metadata says so 

ereslibre:
Fixes: https://github.com/kubewarden/kwctl/issues/40
Depends on: https://github.com/kubewarden/policy-evaluator/pull/11


ereslibre:
Ugh, got the wrong source (the fork). Reopening.

Issue #52, opened 2021-06-23
Add shell completions


flavio:
Add bash, zsh,... completions for kwctl.

The clap crate, which we use to build the cli, has a built-in functionality that generates the completion files.

PR #53, opened 2021-06-23
Completions

flavio:
Add `completions` sub-command. This will generate autocompletion scripts for the major shells, courtesy of clap.

I've also added some instructions to the README about how to use these completion scripts


flavio:
Fixes https://github.com/kubewarden/kwctl/issues/52

ereslibre (approved):
I love it! 👏

PR #56, opened 2021-07-20
Initial integration for running OPA/OPA Gatekeeper policies

ereslibre:
This is a super initial integration. Since we haven't fully discussed the metadata to use to decide the runtime, a commit is added that forces `Opa` for testing manually. However, this needs to change before this PR is merged.

Fixes: https://github.com/kubewarden/policy-evaluator/issues/15

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -3,7 +3,7 @@ use kube::Client;
 use policy_evaluator::{
     cluster_context::ClusterContext,
     constants::*,
-    policy_evaluator::{PolicyEvaluator, ValidateRequest},
+    policy_evaluator::{PolicyEvaluator, PolicyExecutionMode, ValidateRequest},
     policy_metadata::Metadata,
 };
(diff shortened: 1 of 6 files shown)


src/run.rs
```diff
@@ -50,6 +50,7 @@ pub(crate) async fn pull_and_run(
     let policy_evaluator = PolicyEvaluator::from_file(
         policy_id,
         policy_path,
+        PolicyExecutionMode::Opa,
```
ereslibre, line 53: This is here just to try if you want. We have to discuss whether we want to add a flag besides the metadata that we are going to read in order to use one method of evaluation or another.

flavio:
This is good to get everything started, let's keep everything inside of this branch. We will work on the PR once everything is ready

flavio (changes requested):
LGTM, just merge it into a new branch named `opa`

Issue #57, opened 2021-07-21
Allow to annotate policies with a specific runtime


ereslibre:
Depends on: https://github.com/kubewarden/policy-evaluator/issues/21

Allow to annotate policies with a specific execution model, of the three:

- Kubewarden + waPC
- OPA
- OPA Gatekeeper

This information has to be provided as part of the metadata in the `--metadata-path` argument.

flavio:
I would also add that `annotate` should do a quick heuristic against the policy to understand whether this is a Rego vs Kubewarden policy. If the user tries to annotate a Kubewarden policy as OPA/Gatekepper, the annotate command should exit with an error.

The heuristic can leverage these facts:
  * Rego based policies have the OPA ABI constants defined
  * Kubewarden policies must implement the policy protocol guest call

flavio:
Implemented via https://github.com/kubewarden/kwctl/pull/60

PR #60, opened 2021-08-13
Handle execution mode inside of annote and inspect

flavio:
Update the `annotate` and `inspect` commands to handle the new `executionMode` metadata attribute.

This fixes https://github.com/kubewarden/kwctl/issues/57

### DIFF
--- a/src/annotate.rs
+++ b/src/annotate.rs
@@ -1,10 +1,7 @@
+use crate::backend::{Backend, BackendDetector};
 use anyhow::{anyhow, Result};
 use kubewarden_policy_sdk::metadata::ProtocolVersion;
-use policy_evaluator::{
-    constants::*,
-    policy_evaluator::{PolicyEvaluator, PolicyExecutionMode},
-    policy_metadata::Metadata,
-};
+use policy_evaluator::{constants::*, policy_metadata::Metadata};
 use std::fs::File;
 use std::path::PathBuf;
 use validator::Validate;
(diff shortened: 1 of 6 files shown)


src/annotate.rs
```diff
@@ -30,17 +35,76 @@ fn protocol_detector(wasm_path: PathBuf) -> Result<ProtocolVersion> {
...
+
+    Ok(false)
+}
+
 fn prepare_metadata(
     wasm_path: PathBuf,
     metadata_path: PathBuf,
-    detect_protocol_func: impl Fn(PathBuf) -> Result<ProtocolVersion>,
+    kubewarden_protocol_detector_func: impl Fn(PathBuf) -> Result<ProtocolVersion>,
+    rego_detector_func: impl Fn(PathBuf) -> Result<bool>,
```
ereslibre, line 63: I think I prefer to shuffle this data structures a bit, such as:

```rust
enum Backend {
  Unknown,
  OPA,
  OPAGatekeeper,
  KubewardenWapc(ProtocolVersion),
}
```

And we can have a function that performs the heuristics and return a `Result<Backend>`. By decomposing the result, we get the relevant information in each case -- right now, the protocol version in the case of the kubewarden protocol enum value.

The mocking strategy is similar to this one, in that we can pass this heuristics function as an `impl Fn` argument we can mock.

ereslibre (commented):
LGTM, some comments regarding some type refactor that I think would be nice.

src/annotate.rs
flavio: I like the proposal, I'll change the code to reflect that. Thanks!

flavio:
@ereslibre I've just pushed a commit which implements the `Backend` enum and the `BackedDetector` struct.

Once you're fine with it, I'll squash the commits around and merge it

ereslibre (approved):
Thanks a lot, it looks really good :)

PR #72, opened 2021-10-12
Do not use clap macros

flavio:
Switch to "traditional" approach. Some features are not exposed by the macros, such as:

  * Allowed values
  * `number_of_values": this is going to be required to implement the
  `verify` subcommand


flavio:
Ouch, I'll fix the typo and merge it

Issue #66, opened 2021-09-17
kwctl pull: omitting the tag leads to a wrong policy file to be saved


flavio:
Pulling a kwctl policy without specifying the tag will pull the `:latest` tag (which is good), but will save an untagged policy locally.

# How to reproduce

```console
$ kwctl pull registry://ghcr.io/kubewarden/policies/psp-capabilities
$ kwctl policies
+---------------------------------------------------------------------------------+----------+---------------+--------------+-----------+
| Policy                                                                          | Mutating | Context aware | SHA-256      | Size      |
+---------------------------------------------------------------------------------+----------+---------------+--------------+-----------+
| registry://ghcr.io/kubewarden/policies/psp-capabilities                         | yes      | no            | 78fc9dd83ff6 | 1.54 MB   |
```

As you can see the policy doesn't have the tag specified.

Moreover, explicitly pulling the `latest` tag leads to this situation:

```console
$ kwctl pull registry://ghcr.io/kubewarden/policies/psp-capabilities:latest
$ kwctl policies                                                           
+---------------------------------------------------------------------------------+----------+---------------+--------------+-----------+
| Policy                                                                          | Mutating | Context aware | SHA-256      | Size      |
+---------------------------------------------------------------------------------+----------+---------------+--------------+-----------+
| registry://ghcr.io/kubewarden/policies/psp-capabilities                         | yes      | no            | 78fc9dd83ff6 | 1.54 MB   |
| registry://ghcr.io/kubewarden/policies/psp-capabilities:latest                  | yes      | no            | 78fc9dd83ff6 | 1.54 MB   |
```
# Desired outcome

Executing the following command:

```console
$ kwctl pull registry://ghcr.io/kubewarden/policies/psp-capabilities
```
pulls the `:latest` tag of the policy, and crates the following file

```console
$ kwctl policies                                                           
+---------------------------------------------------------------------------------+----------+---------------+--------------+-----------+
| Policy                                                                          | Mutating | Context aware | SHA-256      | Size      |
+---------------------------------------------------------------------------------+----------+---------------+--------------+-----------+
| registry://ghcr.io/kubewarden/policies/psp-capabilities:latest                  | yes      | no            | 78fc9dd83ff6 | 1.54 MB   |
```

PR #76, opened 2021-10-19
Add latest tag if not present when pulling from a registry

raulcabello:
fixes  kubewarden/kwctl#66


ereslibre:
I would recommend to address this logic in the `policy-fetcher` though. This way, the `policy-server` will also benefit from this logic.

I think we can address this logic at https://github.com/kubewarden/policy-fetcher/blob/2da685fbb123986838357bffcbab5f0c48e23b88/src/lib.rs#L60, if the scheme is `registry` here, and the tag is empty, then default directly to `:latest`.

PR #78, opened 2021-10-21
Implement the verify sub-command

flavio:
This command verifies the integrity of a policy stored inside of an OCI registry using Sigstore.

If the policy has been downloaded, the code will also ensures nobody tampered the local file.


flavio:
I'll update Cargo.toml once we merge and tag a new release of policy-fetcher

PR #83, opened 2021-11-04
Bump policy-fetcher and update existing code

ereslibre:


viccuad (approved):

viccuad:
I don't understand why the e2e tests fail, though.

ereslibre:
> I didn't look at why the e2e tests fail, though. Change in path on output?

I will try to check before leaving today, but I fear it has to do with bats reporting error if stderr is not empty.

ereslibre:
NVM, it happens because the test is looking for a string that is not in stdout anymore since https://github.com/kubewarden/policy-fetcher/pull/20

viccuad:
> NVM, it happens because the test is looking for a string that is not in stdout anymore since [kubewarden/policy-fetcher#20](https://github.com/kubewarden/policy-fetcher/pull/20)

Could we enable merging only after the ci is green? (I don't have perms). Also, shouldn't the security audit reactive been run, as cargo.{toml, lock} was touched?

ereslibre:
> Could we enable merging only after the ci is green? (I don't have perms)

Done.

> Also, shouldn't the security audit reactive been run, as cargo.{toml, lock} was touched?

I am really low on GitHub actions knowledge, but I believe the yaml is missing (pseudocode) `on: pull_request`?

viccuad:
> I am really low on GitHub actions knowledge, but I believe the yaml is missing (pseudocode) `on: pull_request`?

Could be. But other PRs seem to have run it: https://github.com/kubewarden/kwctl/pull/79. Not that it matters much, we can have an eye on it in the future.

viccuad:
From https://github.com/actions-rs/audit-check:
> Due to token permissions, this Action WILL NOT be able to create Checks for Pull Requests from the forked repositories, see actions-rs/clippy-check#2 for details.

So I suppose that's that. There's no problem, as it runs on main.

PR #84, opened 2021-11-12
feat!: Add sigstore verification to `kwctl {pull, run}`

viccuad:
`kwctl {pull, run, verify}` now accept new flags. The flags are similar to cosign et al flags, and multiple are allowed:
* `-a`, `--verification-annotation`: also `-a` in `cosign`.
* `-k`, `verification-key`: corresponds to  `-key` in `cosign`. Clap only supports POSIX style short flags, not short flags that are more than 1 character.

Now:
- `kwctl pull -k key.pub -a env=prod ...`  will verify prior to pulling, and then verify the local checksum against the manifest.
- `kwctl run -k key.pub -a env=prod ...`  with keys and annotations will verify prior to running the policy.
- `kwctl verify -k key.pub -a env=prod ...`  with keys and annotations will verify the policy.

Breaking change: `kwctl verify` flags have also changed for consistency.

Note: `kwctl {pul, run, verify}` only works for `registry://` protocol if passing keys, as the rest of verification features.

Note:  When missing annotations on verification, we fail with a bogus string. It should return instead: `Annotation not satisfied missing_annotation="stable"`. This needs to be changed in sigstore-rs.
For that, we need to consume sigstore-rs-v0.1.0, which expects a rekor key for building the client. I wonder if we want to ask the users to pull a rekor key for verifying something, regarless of rekor functionality. I suggest gating rekor functionality under an experimental env var, as done in cosign.

Note: Refactoring calls to verify() so it uses references instead of cloning objects is not done yet, as it needs changes in sigstore-rs for `annotations`, and then to consume them in policy-fetcher.


viccuad:
Right now the e2e tests for verification work against the image in `ghcr.io/viccuad/pod-privileged:viccuad` (see e2e-tests/sigstore/README.md). I suppose it would be a good idea to create and use a namespace under the Kubewarden org, such as  `ghcr.io/kubewarden/tests/pod-privileged:v0.1.9`. But I lack rights for that.

viccuad:
As a curiosity, one can see the [signed manifest of the image under test on GH](https://github.com/orgs/kubewarden/packages/container/tests%2Fpod-privileged/10198620?tag=sha256-0d6611ea12cf2904066308dde1c480b5d4f40e19b12f51f101a256b44d6c2dd5.sig).

src/main.rs
```diff
@@ -78,36 +79,54 @@ async fn main() -> Result<()> {
                     None => PullDestination::MainStore,
                 };
                 let (sources, docker_config) = remote_server_options(matches)?;
-                pull::pull(uri, docker_config, sources, destination).await?;
+                let (key_files, annotations) = verification_options(matches)?;
+
+                // verify policy prior to pulling if keys listed, and keep the
+                // verified manifest digest of last iteration, even if all are
+                // the same:
+                let mut verified_manifest_digest: String = String::new();
```
flavio, line 87: A better way is to use an `Option`:

```suggestion
let mut verified_manifest_digest: Option<String> = None;
```

src/main.rs
```diff
@@ -78,36 +79,54 @@ async fn main() -> Result<()> {
...
+                for key_file in key_files.iter() {
+                    // let key = verify::read_key_file(&key_file[0])?;
+                    verified_manifest_digest = verify::verify(
+                        uri,
+                        docker_config.clone(),
+                        sources.clone(),
+                        annotations.clone(),
```
flavio, line 94: I would pass all of them by reference. It's up to `verify:verify` to clone them if needed

src/main.rs
```diff
@@ -78,36 +79,54 @@ async fn main() -> Result<()> {
...
+                        &key_file[0],
+                    )
+                    .await?;
```
flavio, line 97: Now that `verified_manifest_digest` is an `Option`, you just have to change this line to `await?.ok();`

src/main.rs
```diff
@@ -78,36 +79,54 @@ async fn main() -> Result<()> {
...
                 let (sources, docker_config) = remote_server_options(matches)?;
-                pull::pull(uri, docker_config, sources, destination).await?;
+                let (key_files, annotations) = verification_options(matches)?;
+
+                // verify policy prior to pulling if keys listed, and keep the
+                // verified manifest digest of last iteration, even if all are
+                // the same:
+                let mut verified_manifest_digest: String = String::new();
+                for key_file in key_files.iter() {
+                    // let key = verify::read_key_file(&key_file[0])?;
```
flavio, line 89: This can be dropped

src/main.rs
```diff
@@ -78,36 +79,54 @@ async fn main() -> Result<()> {
...
-                        } else {
-                            Some(values)
-                        }
-                    }
-                };
-
-                verify::verify(uri, docker_config, sources, annotations, key_file).await?;
+                for key_file in key_files.iter() {
+                    println!("{:}", &key_file[0]);
+                    // let key = verify::read_key_file(&key_file[0])?;
```
flavio, line 120: this can be removed

src/main.rs
```diff
@@ -78,36 +79,54 @@ async fn main() -> Result<()> {
...
+                    verify::verify(
+                        uri,
+                        docker_config.clone(),
+                        sources.clone(),
+                        annotations.clone(),
```
flavio, line 125: Same here, I would pass all of them by reference

src/main.rs
```diff
@@ -193,13 +212,31 @@ async fn main() -> Result<()> {
...
                         None
                     };
+                let (key_files, annotations) = verification_options(matches)?;
+
+                // verify policy prior to pulling if keys listed, and keep the
+                // verified manifest digest of last iteration, even if all are
+                // the same:
+                let mut verified_manifest_digest: String = String::new();
+                for key_file in key_files.iter() {
+                    // let key = verify::read_key_file(&key_file[0])?;
```
flavio, line 222: to be removed

flavio (changes requested):
Overall looks good to me. I've left some comments, some of the changes requested are caused by "mistakes" we have done in our early days of Rust... I can help to address them

flavio:
BTW, I'm about to push out a new iteration of `sigstore-rs` that will support verification of artifacts signed in keyless mode. This will require the end-users to provide not only rekor's public key, but also fulcio's certificate.

cosign has two ways to obtain these information:
  1. Rely on some hard coded data
  2. Fetch the data from Sigstore's TUF repository 

Unfortunately the 2nd option is not feasible yet, because of a bug I'm currently working on (see https://github.com/sigstore/sigstore-rs/issues/9).

In the meantime I would go ahead and implement the 1st solution

src/main.rs
viccuad: Indeed, thanks! Changed it in the last fixup commits; I can rebase after the review (I wonder what is preferable, to re-review it all, or for me to push fixup commits that will dissapear with a rebase after the review).

src/main.rs
viccuad: Fixed, thanks!

src/main.rs
viccuad: Fixed, thanks!

src/main.rs
viccuad: Fixed, thanks!

src/main.rs
viccuad: Fixed, thanks!

src/main.rs
viccuad: Fixed, thanks!

src/main.rs
viccuad: Fixed, thanks!

src/verify.rs
```diff
@@ -5,25 +5,52 @@ use policy_fetcher::verify::Verifier;
...
+pub(crate) type VerificationAnnotations = HashMap<String, String>;
+
 pub(crate) async fn verify(
     url: &str,
-    docker_config: Option<DockerConfig>,
-    sources: Option<Sources>,
-    annotations: Option<HashMap<String, String>>,
+    docker_config: &Option<DockerConfig>,
+    sources: &Option<Sources>,
+    annotations: &Option<VerificationAnnotations>,
```
flavio, line 14: Move the reference inside of the `Option`. You can then turn something like `Option<DockerConfig>` into `Option<&DockerConfig>` by doing:

```rust
let docker_config: Option<DockerConfig> = Some(DockerConfig::default());
let dc = Option<&DockerConfig> = docker_config.as_ref();
```

src/verify.rs
```diff
@@ -5,25 +5,52 @@ use policy_fetcher::verify::Verifier;
...
+}
+
+fn read_key_file(path: &str) -> Result<String> {
     let mut pub_key_file =
-        File::open(key_file).map_err(|e| anyhow!("Cannot read verification key file: {:?}", e))?;
+        File::open(path).map_err(|e| anyhow!("Cannot read verification key file: {:?}", e))?;
     let mut verification_key = String::new();
     pub_key_file
         .read_to_string(&mut verification_key)
         .map_err(|e| anyhow!("Error reading contents of verification key file {:?}", e))?;
```
flavio, line 53: Something from the old days, not your fault. This can be done in a more succinct way:

```rust
let contents = std::fs::read_to_string("/etc/motd")
  .map_err(|e| anyhow!("Something went wrong: {:?}", e)?;
```

src/verify.rs
```diff
@@ -5,25 +5,52 @@ use policy_fetcher::verify::Verifier;
...
+    docker_config: &Option<DockerConfig>,
+    sources: &Option<Sources>,
+    verified_manifest_digest: &str,
 ) -> Result<()> {
+    let mut verifier = Verifier::new(sources.clone());
+    verifier
+        .verify_local_file_checksum(url, docker_config.clone(), verified_manifest_digest)
+        .await?;
+
+    println!("Local checksum successfully verified");
```
flavio, line 43: I think we should uses tracing.info for that.

src/verify.rs
```diff
@@ -5,25 +5,52 @@ use policy_fetcher::verify::Verifier;
...
+    let verified_manifest_digest = verifier
+        .verify(
+            url,
+            docker_config.clone(),
+            annotations.clone(),
+            &verification_key,
+        )
+        .await?;
+
+    println!("Policy successfully verified");
```
flavio, line 28: I think we should uses tracing.info for that.

flavio (changes requested):
Great improvements, I left some more comments

src/verify.rs
viccuad: Done in https://github.com/kubewarden/kwctl/pull/84/commits/595796007840a6eb43ffd9bc95a004aebe3e98ed.
I wonder if we should print the info logs differently (without the timestamped prefix `Nov 16 12:21:56.483  INFO`) if used in an interactive console.

src/verify.rs
viccuad: Done in https://github.com/kubewarden/kwctl/pull/84/commits/595796007840a6eb43ffd9bc95a004aebe3e98ed.

viccuad:
Updated policy-fetcher to released v0.1.9 instead of my fork. Will rebase autosquash when it's approved.

src/verify.rs
viccuad: True, thanks! done in https://github.com/kubewarden/kwctl/pull/84/commits/595796007840a6eb43ffd9bc95a004aebe3e98ed

src/verify.rs
flavio: Yes, I would tackle that step by step. Maybe once we're done with this change and my work on the "get policy manifest" callback we can make a single PR that changes all of that.

Does it make sense?

viccuad: Yes, thanks for the clarification!

Done in https://github.com/kubewarden/kwctl/pull/84/commits/ebee37c89eb96ab442e2b13c5267829e3fdbd93e.

flavio (approved):
Great job :clap: :ship:

viccuad:
rebased, ready to merge.

PR #85, opened 2021-11-16
Enable "get OCI manifest digest" callback

flavio:
Ensure the Kubewarden policies executed via kwctl can request the host to fetch the manifest digest of an OCI image.

Note well: the current code works, but I wonder what is the right place of the new `src/callback_handler` directory should be. This is code that is probably going to be shared with the policy-server as it currently is.
I need to do more thinking about that.


flavio:
There are still some minor details to sort out, but this is now ready for review.

ereslibre (approved):
LGTM modulo lints

viccuad (approved):
LGTM. Maybe it can be a module in policy-evaluator?

flavio:
> LGTM. Maybe it can be a module in policy-evaluator?

Yes, either a module or, even better, a crate that lives in there. What do you think folks?

ereslibre:
> Yes, either a module or, even better, a crate that lives in there. What do you think folks?

I'm good with both. As we depend on policy-evaluator already so I would favor a module at this point, and create a crate out of it if we see the need.

flavio:
This is now done via https://github.com/kubewarden/kwctl/issues/205

PR #87, opened 2021-11-17
Update CHANGELOG.md

viccuad:
Just a run of `git-chglog --next-tag v0.2.3 --output CHANGELOG.md`.
I don't have push perms to main.
We can tag the new release once this is in main.
Note: The follow-up release+1 would list the PR bumping the changelog from this v0.2.3 release as part of release+1 contents

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,30 @@
 ## [Unreleased]
 
 
+<a name="v0.2.3"></a>
+## [v0.2.3] - 2021-11-16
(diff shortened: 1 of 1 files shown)


ereslibre (approved):
LGTM, I think we have to document how to create a new release so it's written down

viccuad:
> LGTM, I think we have to document how to create a new release so it's written down

Indeed! done in https://github.com/kubewarden/kwctl/pull/88.

PR #89, opened 2021-11-17
Fix tokio CVE issue

flavio:
Ensure patched version of tokio is being used.

This cannot be merged until a new version of policy-evaluator is published, since both kwctl and policy-fetcher must use the same version of the `kube` crate.


ereslibre:
Superseded by https://github.com/kubewarden/kwctl/pull/90

PR #91, opened 2021-11-19
tests: Expand `kwctl run` cases, test `kwctl verify`

viccuad:
Before, the "verify a signed policy from an OCI registry" test was
calling `kwctl pull`, not `kwctl verify`. No harm done, as the same
codepath is executed since `kwctl pull` calls verify::verify().

Add case for `kwctl run` when verification is incorrect.

### DIFF
--- a/e2e-tests/e2e.bats
+++ b/e2e-tests/e2e.bats
@@ -82,32 +82,40 @@ kwctl() {
 }
 
 @test "verify a signed policy from an OCI registry" {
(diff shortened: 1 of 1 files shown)


flavio (changes requested):
LGTM, but one of the e2e tests is failing

viccuad:
Superseded by https://github.com/kubewarden/kwctl/pull/92.

PR #92, opened 2021-11-22
fix: Verification of multiple keys

viccuad:
Fix bug when verifying multiple keys in `kwctl {verify, pull, run}`.
Add and fix tests to that effect.
Re-sign image under test.


flavio (approved):
Everything looks good to me

PR #95, opened 2021-12-01
replace openssl with rustls

ereslibre:


ereslibre:
Cannot request a review from @kubewarden/kubewarden-developers proper...

Cargo.toml
```diff
@@ -16,12 +16,12 @@ clap = "2.33.3"
...
-kube = "0.64.0"
+kube = { version = "0.64.0", default-features = false, features = ["client", "rustls-tls"] }
 kubewarden-policy-sdk = "0.2.3"
 lazy_static = "1.4.0"
-mdcat = "0.22"
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.2.4" }
-policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.1.19" }
+mdcat = "0.24"
+policy-evaluator = { git = "https://github.com/ereslibre/policy-evaluator", branch = "rustls-tls" }
+policy-fetcher = { git = "https://github.com/ereslibre/policy-fetcher", branch = "rustls-tls" }
```
jvanz, line 24: Use the `kubewarden` repos?

ereslibre: Same reason as https://github.com/kubewarden/policy-server/pull/131/files#r761868059

src/manifest.rs
```diff
@@ -72,7 +72,8 @@ pub(crate) fn manifest(uri: &str, resource_type: &str, settings: Option<String>)
     let resource = match resource_type {
         "ClusterAdmissionPolicy" => ClusterAdmissionPolicy::try_from(scaffold_data),
         _other => Err(anyhow!(
-            "Resource {} unknown. Valid types are: ClusterAdmissionPolicy"
+            "Resource {} unknown. Valid types are: ClusterAdmissionPolicy",
+            _other,
```
flavio, line 76: silly question, given we're now using `_other`, shouldn't we change its name to be `other`?

flavio (approved):
LGTM

src/manifest.rs
ereslibre: Absolutely. Thanks!

flavio (approved):
LGTM

ereslibre:
I'm updating this PR so it reads the fulcio cert and the rekor public key as an argument -- or defaults to a location in ~.

src/main.rs
```diff
@@ -417,3 +444,37 @@ fn verification_options(
...
+            None
+        };
+
+    if fulcio_cert.is_none() && rekor_public_key.is_none() {
+        return Ok(None);
+    }
+
+    if fulcio_cert.is_none() || rekor_public_key.is_none() {
+        return Err(anyhow!(
+            "both a fulcio certificate and a rekor public key are required"
```
flavio, line 472: I would expand that to give a hint about how to fix this problem, something like:

```suggestion
            "both a fulcio certificate and a rekor public key are required, these can be obtained by using the `cosign initialize` command"
```

src/run.rs
```diff
@@ -19,6 +19,7 @@ pub(crate) async fn pull_and_run(
     request: &str,
     settings: Option<String>,
     verified_manifest_digest: &Option<String>,
+    sigstore_opts: &SigstoreOpts,
```
flavio, line 22: clippy is complaining about the number of args we are passing. Maybe this is something we can fix by aggregating some of them into ad-hoc structs?

src/verify.rs
```diff
@@ -46,7 +49,5 @@ pub(crate) async fn verify_local_checksum(
 }
 
 fn read_key_file(path: &str) -> Result<String> {
-    let verification_key =
-        fs::read_to_string(path).map_err(|e| anyhow!("Something went wrong: {:?}", e))?;
-    Ok(verification_key)
+    fs::read_to_string(path).map_err(|e| anyhow!("could not read file {}: {:?}", path, e))
```
flavio, line 52: Maybe this could be done inside of the constructor of `SigstoreOpts`?

flavio (commented):
Overall great, I left some comments

src/run.rs
ereslibre: I already did on this argument (it's a single struct containing two arguments). I also didn't want to block longer than needed for that refactor. I think the argument refactoring can be done later.

If you agree with that I'm taking care of the refactor, but we'd have to merge with a red clippy. I just didn't want to get into any parameter refactoring on this PR, that we are waiting for.

flavio: Yes, that makes total sense to me!

src/verify.rs
```diff
@@ -13,16 +13,17 @@ pub(crate) async fn verify(
     sources: Option<&Sources>,
     annotations: Option<&VerificationAnnotations>,
     key_file: &str,
+    fulcio_cert: &[u8],
+    rekor_public_key: &str,
```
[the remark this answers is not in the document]
ereslibre: I would leave this to the refactor that I will do after this PR

src/verify.rs
ereslibre: This is about the verification key though (an argument that was present already), not directly related to the new fulcio & rekor new arguments.

src/backend.rs
```diff
@@ -37,13 +37,10 @@ fn rego_policy_detector(wasm_path: PathBuf) -> Result<bool> {
...
-        String::from(wasm_path.to_string_lossy()),
-        wasm_path.as_path(),
-        PolicyExecutionMode::KubewardenWapc,
-        None,
-    )?;
-    policy_evaluator
+    PolicyEvaluatorBuilder::new("".to_string())
+        .policy_file(&wasm_path)?
+        .execution_mode(PolicyExecutionMode::KubewardenWapc)
+        .build()?
```
flavio, line 43: Love it!

src/run.rs
```diff
@@ -32,7 +36,15 @@ pub(crate) async fn pull_and_run(
...
 
     if let Some(digest) = verified_manifest_digest {
-        verify::verify_local_checksum(&uri, docker_config, sources, digest).await?
+        verify::verify_local_checksum(
+            &uri,
+            docker_config,
+            sources,
+            digest,
+            &sigstore_opts.fulcio_cert,
+            &sigstore_opts.rekor_public_key,
```
flavio, line 45: What I meant here was to use `&sigstore_opts` as a param. We can change that later if we really want to

flavio (approved):
LGTM

src/run.rs
ereslibre: Ah sorry, right. I misunderstood you.

viccuad (approved):

jvanz (approved):

ereslibre:
Converting to draft, given e2e tests related to sigstore are removed temporarily.

flavio (approved):
LGTM

viccuad (approved):

PR #98, opened 2021-12-23
Build kwctl statically for different architectures and OS's

ereslibre:


ereslibre:
Cross compilation to Windows using the `msvc` stack (`x86_64-pc-windows-msvc`) is not working due to https://github.com/rust-lang/stacker/issues/58.

The approach described in https://jake-shadle.github.io/xwin/ worked to a big extent until I hit the error described in the issue.

Namely:

```
     Running `rustc --crate-name serde /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.132/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts --crate-type lib --emit=dep-info,metadata,link -C opt-level=3 -C embed-bitcode=no --cfg 'feature="default"' --cfg 'feature="derive"' --cfg 'feature="serde_derive"' --cfg 'feature="std"' -C metadata=7c40528e715598af -C extra-filename=-7c40528e715598af --out-dir /kwctl/target/x86_64-pc-windows-msvc/release/deps --target x86_64-pc-windows-msvc -C linker=lld-link -L dependency=/kwctl/target/x86_64-pc-windows-msvc/release/deps -L dependency=/kwctl/target/release/deps --extern serde_derive=/kwctl/target/release/deps/libserde_derive-d37a830ac08993b6.so --cap-lints allow -Lnative=/xwin/crt/lib/x86_64 -Lnative=/xwin/sdk/lib/um/x86_64 -Lnative=/xwin/sdk/lib/ucrt/x86_64 --cfg ops_bound --cfg core_reverse --cfg de_boxed_c_str --cfg de_boxed_path --cfg de_rc_dst --cfg core_duration --cfg integer128 --cfg range_inclusive --cfg iterator_try_fold --cfg num_nonzero --cfg serde_derive --cfg core_try_from --cfg num_nonzero_signed --cfg systemtime_checked_add --cfg std_atomic64 --cfg std_atomic`
error: failed to run custom build command for `psm v0.1.16`

Caused by:
  process didn't exit successfully: `/kwctl/target/release/build/psm-3a8e7e6c1a830ca4/build-script-build` (exit status: 1)
  --- stdout
  OPT_LEVEL = Some("3")
  TARGET = Some("x86_64-pc-windows-msvc")
  HOST = Some("x86_64-unknown-linux-gnu")
  CC_x86_64-pc-windows-msvc = None
  CC_x86_64_pc_windows_msvc = Some("clang-cl")
  CFLAGS_x86_64-pc-windows-msvc = None
  CFLAGS_x86_64_pc_windows_msvc = None
  TARGET_CFLAGS = None
  CFLAGS = None
  CRATE_CC_NO_DEFAULTS = None
  CARGO_CFG_TARGET_FEATURE = Some("fxsr,sse,sse2")
  DEBUG = Some("false")
  cargo:rustc-cfg=asm
  running: "ml64.exe" "-nologo" "-Fo/kwctl/target/x86_64-pc-windows-msvc/release/build/psm-ca826a11c832451b/out/src/arch/x86_64_msvc.o" "-c" "src/arch/x86_64_msvc.asm"

  --- stderr


  error occurred: Failed to find tool. Is `ml64.exe` installed?


warning: build failed, waiting for other jobs to finish...
error: build failed
```

ereslibre:
I could build a Windows executable with the msvc toolchain just fine.

.github/workflows/release.yml
```diff
@@ -1,15 +1,7 @@
 name: kwctl release
-
-on:
-  push:
-    tags:
-    - 'v*'
```
flavio, line 6: I guess you removed these constrains to debug building inside of your fork. We should bring them back now

.github/workflows/release.yml
```diff
@@ -18,126 +10,98 @@ jobs:
...
+  #   - name: Create Release
+  #     id: create-release
+  #     uses: actions/create-release@v1
+  #     env:
+  #       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  #     with:
+  #       tag_name: ${{ github.ref }}
+  #       release_name: Release kwctl ${{ github.ref }}
+  #       draft: false
+  #       prerelease: false
```
flavio, line 33: This should be uncommented

.github/workflows/release.yml
```diff
@@ -18,126 +10,98 @@ jobs:
...
+      env:
+        CC: x86_64-linux-musl-gcc
+      run: cargo build --target=x86_64-unknown-linux-musl --release
+    # - name: Upload Release Asset
+    #   uses: actions/upload-release-asset@v1
+    #   env:
+    #     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    #   with:
+    #     upload_url: ${{ jobs.release.steps.create-release.outputs.upload_url }}
+    #     asset_name: kwctl-linux-x86_64
```
flavio, line 60: Same here, should be brought back

.github/workflows/release.yml
```diff
@@ -18,126 +10,98 @@ jobs:
...
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        CC: aarch64-linux-musl-gcc
+      run: cargo build --target=aarch64-unknown-linux-musl --release
+    # - name: Upload Release Asset
+    #   uses: actions/upload-release-asset@v1
+    #   env:
+    #     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    #   with:
+    #     upload_url: ${{ jobs.release.steps.create-release.outputs.upload_url }}
```
flavio, line 86: Same here, should be brought back

.github/workflows/release.yml
```diff
@@ -18,126 +10,98 @@ jobs:
...
           args: --workspace
-
-  build:
-    name: Build Release
-    needs: test
+  # release:
+  #   name: Create release
+  #   runs-on: ubuntu-latest
+  #   needs:
+  #     - test
```
flavio, line 22: I think we should change it so that `release` requires `test` **and** `build` steps. The `build` steps would require the `test` ones.

.github/workflows/release.yml
```diff
@@ -18,126 +10,98 @@ jobs:
...
+  #     with:
+  #       tag_name: ${{ github.ref }}
+  #       release_name: Release kwctl ${{ github.ref }}
+  #       draft: false
+  #       prerelease: false
+  build-linux-x86_64:
+    name: Build x86_64 (musl) binary
     runs-on: ubuntu-latest
+    # needs:
+    #   - release
```
flavio, line 38: Should require `test`

.github/workflows/release.yml
```diff
@@ -18,126 +10,98 @@ jobs:
...
+    #   env:
+    #     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    #   with:
+    #     upload_url: ${{ jobs.release.steps.create-release.outputs.upload_url }}
+    #     asset_name: kwctl-linux-x86_64
+  build-linux-aarch64:
+    name: Build aarch64 (musl) binary
     runs-on: ubuntu-latest
+    # needs:
+    #   - release
```
flavio, line 65: Should require `test`

.github/workflows/release.yml
```diff
@@ -18,126 +10,98 @@ jobs:
...
+    #   uses: actions/upload-release-asset@v1
+    #   env:
+    #     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    #   with:
+    #     upload_url: ${{ jobs.release.steps.create-release.outputs.upload_url }}
+  build-windows-x86_64:
+    name: Build x86_64 (msvc) binary
+    runs-on: windows-latest
+    # needs:
+    #   - release
```
flavio, line 91: Should require `test`

.github/workflows/release.yml
```diff
@@ -18,126 +10,98 @@ jobs:
...
+    - run: rustup target add x86_64-pc-windows-msvc
+    - name: Build kwctl
+      run: cargo build --target=x86_64-pc-windows-msvc --release
+    # - name: Upload Release Asset
+    #   uses: actions/upload-release-asset@v1
+    #   env:
+    #     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    #   with:
+    #     upload_url: ${{ jobs.release.steps.create-release.outputs.upload_url }}
+    #     asset_name: kwctl-windows-x86_64
```
flavio, line 107: Should be brought back

flavio (changes requested):
Overall great. I think we should build `kwctl` only when a new tag is created.

When a new tag is created we should:
  1. run the units tests
  2. build the `kwctl` binary for all the architectures 
  3. create a GH Release
  4. Upload the binaries to the release

Also, it would be nice to run also the e2e tests in between step 2 and 3, but maybe this could be done inside of another PR

flavio:
> I could build a Windows executable with the msvc toolchain just fine.

The real question is: does it actually work? ;)

Do you think it would be possible to run the e2e tests on windows too?

.github/workflows/release.yml
ereslibre: I was testing the GitHub actions in my own fork. I should probably have used another branch.

.github/workflows/release.yml
ereslibre: Same.

.github/workflows/release.yml
ereslibre: Same.

.github/workflows/release.yml
ereslibre: I think that makes sense. I will refactor the GitHub actions a little, so that:

- Every build step depends on tests. Upon success, upload binary artifact for that run.
- Create release step (that depends on all build steps). This:
  - Creates the release on GitHub
  - Populates every binary by downloading every artifact (every binary for different archs) and uploading it as a release asset.

ereslibre:
> The real question is: does it actually work? ;)

It doesn't. It creates an empty file in the store when trying to pull a policy.

I think is better to not block on that. My approach would be to publish all assets and work on the fix for the Windows binary afterwards. The fact that we have a Windows binary is already progress.

> Do you think it would be possible to run the e2e tests on windows too?

I think we should do it, yes. I would not do that right now due to the problem I mentioned.

Issue #97, opened 2021-12-15
[Feature request] Use SHA value from `kwctl` 


nunix:
[Actual behaviour]
For commands like `inspect` or `manifest`, `kwctl` needs the policy "name", which currently is the <scheme>://<path to policy>

This can bring some confusion, specially when working with the `registry` scheme, as it can lead someone to think that it can be used "online" (read: without pulling the image).

[Desired behaviour]
Like container images, the policies also have a "hash", so it could be used for working with it locally.
Example:
```shell
$ kwctl policies
+--------------------------------------------------------------+----------+---------------+----------------+----------+
| Policy                                                                                                 | Mutating  | Context aware | SHA-256            | Size          |
+--------------------------------------------------------------+----------+---------------+----------------+----------+
| registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9      | no             | no                      | 59e34f482b40 | 21.86 kB   |
+--------------------------------------------------------------+----------+---------------+----------------+----------+

$ kwctl inspect 59e34f482b40
Details
title:              pod-privileged
description:        Limit the ability to create privileged containers
author:             Flavio Castelli
url:                https://github.com/kubewarden/pod-privileged-policy
source:             https://github.com/kubewarden/pod-privileged-policy
license:            Apache-2.0
mutating:           false
context aware:      false
execution mode:     kubewarden-wapc
protocol version:   1
...
```

Hope it makes sense and any other way is welcome 😄

flavio:
Makes sense to me

nitishm:
@flavio I took a stab at this. Please let me know if this is what you were thinking or if we want to support a `--sha` flag instead. 
In this implementation we can either pass the URI or the SHA (truncated to 12 characters) and it uses regex to figure out which is which.

PR #99, opened 2021-12-27
(feat) Use SHA value in kwctl manifest lookup

nitishm:
Partially addresses #97 

Support looking up policies using the SHA value of the policy

Signed-off-by: Nitish Malhotra <nitish.malhotra@gmail.com>


flavio:
It looks like a subset of https://github.com/kubewarden/kwctl/pull/100 :thinking: . Can we close it?

nitishm:
Yes it should be safe to do so IMO

PR #100, opened 2021-12-28
(feat) Use sha value for kwctl manifest cmd

### LINKED ISSUE #97: [Feature request] Use SHA value from `kwctl`

nitishm:
Closes #97 

Depends On: #99

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -319,7 +319,7 @@ async fn main() -> Result<()> {
         }
         Some("inspect") => {
             if let Some(matches) = matches.subcommand_matches("inspect") {
-                let uri = matches.value_of("uri").unwrap();
+                let uri = matches.value_of("id").unwrap();
                 let output = inspect::OutputType::try_from(matches.value_of("output"))?;
 
                 inspect::inspect(uri, output)?;
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -1,7 +1,7 @@
 use anyhow::{anyhow, Result};
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
 use serde::{Deserialize, Serialize};
-use std::convert::TryFrom;
+use std::{convert::TryFrom, path::PathBuf};
 use validator::Validate;
 
 use policy_evaluator::policy_metadata::{Metadata, Rule};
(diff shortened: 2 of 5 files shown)


nunix:
Hi @nitishm , just tested it and seems to work just fine (from an usage perspective):
image

I just tried the `inspect` subcommand. And can I kindly ask if it is possible to make it like `docker` cli, that we can use the first 3 characters too (example below)?
image

src/main.rs
```diff
@@ -319,7 +319,7 @@ async fn main() -> Result<()> {
         }
         Some("inspect") => {
             if let Some(matches) = matches.subcommand_matches("inspect") {
-                let uri = matches.value_of("uri").unwrap();
+                let uri = matches.value_of("id").unwrap();
```
flavio, line 322: nitpicking, maybe we should rename the `uri` variable to be `id`, just to be consistent with the cli arg

src/manifest.rs
```diff
@@ -52,23 +52,34 @@ struct ScaffoldData {
...
+    let mut uri: String = id.to_string();
+
+    let is_sha = crate::utils::is_sha(id);
+    if is_sha {
+        uri = crate::utils::uri_from_sha(id)?;
+        wasm_path = crate::utils::wasm_path_from_sha(id)?;
+    } else {
+        let uri = crate::utils::map_path_to_uri(id)?;
+        wasm_path = crate::utils::wasm_path(uri.as_str())?;
+    }
```
flavio, line 66: nitpicking, this code already exists inside of the inspect command. Maybe we could move it into a new dedicated function under `crate::utils`?

src/utils.rs
```diff
@@ -25,6 +31,30 @@ pub(crate) fn map_path_to_uri(uri: &str) -> Result<String> {
     }
 }
 
+pub(crate) fn uri_from_sha(sha: &str) -> Result<String> {
+    let policies = Store::default().list()?;
+    let policy = policies.iter().find(
+        |policy|  {
+            let mut policy_sha = policy.digest().unwrap();
```
flavio, line 38: nitpicking, I would prefer to bubble up the error instead of panicking here.

src/utils.rs
```diff
@@ -25,6 +31,30 @@ pub(crate) fn map_path_to_uri(uri: &str) -> Result<String> {
...
+            policy_sha.truncate(12);
+            policy_sha == *sha
+        })
+        .ok_or_else(|| anyhow!("Cannot find policy by sha '{}' inside of the local store.\nTry executing `kwctl pull on the URI`", sha))?;
+    Ok(policy.local_path.clone())
+}
```
flavio, line 56: The `wasm_path_from_sha` and the `uri_from_sha` looks almost identical.

I think a better approach would be to create a function that, given a shasum, returns the `Policy` object. Maybe this could become a method of the `Store` struct.

flavio (changes requested):
Thanks a lot for your contribution, sorry for the delay caused by the holiday season.

The PR looks good, I left some minor notes.

nitishm:
> Thanks a lot for your contribution, sorry for the delay caused by the holiday season.
> 
> The PR looks good, I left some minor notes.

Excellent. I will address these comments shortly :). And welcome back from the break!

jvanz:
@nitishm are you going to work on this yet?

nitishm:
> @nitishm are you going to work on this yet?

Hey, I dropped the ball on this one because life happened. I can jump on it now or if you'd like to pick it up I'd be happy to hand off. Either works. Lmk

flavio:
@nitishm no worries, we totally get it. If you still want to work on this feature, that would be fine with us!

flavio:
Closing, this has been done via https://github.com/kubewarden/kwctl/pull/587. @nitishm thanks for your initial efforts on this task

PR #103, opened 2022-01-18
 Build kwctl for different architectures and OS's

ereslibre:


jvanz (approved):
LGTM! :D

flavio (changes requested):
From my understanding, whenever we create a tag like v... this action will be triggered. As a first step, we will run the unit tests, then we will create a GH release and then we will build the binaries for linux/win/osx.

If I got this right, that means there could be something going wrong with the windows build (for example), which would lead the release to not have the windows binaries attached to it.

I think it would be better to define the following tasks:
* run tests
* build binaries
* create release

With build binaries depending on run tests, and create release that depends on build binaries.

The build binaries steps can upload the final binaries as artifacts of their run, so that the "create release" task can:
* create the release object
* download the binaries and attach them to the release object
I don't think if the last steps are possible, but I think they should...

ereslibre:
Updated PR:

With this automation (with some minor things updated to match this org instead of my own fork), this tag and artifacts were created on my own fork successfully:

- https://github.com/ereslibre/kwctl/releases (with https://github.com/ereslibre/kwctl/actions/runs/1717769319)

flavio (approved):
LGTM, thanks!

Issue #96, opened 2021-12-14
[Feature request] Set a name with `kwctl manifest`


nunix:
[Actual state]
Currently, when generating a `ClusterAdmissionPolicy` with `kwctl manifest`, the `metadata[].name` value is set to "generated-policy".

Therefore, if we apply the manifest as is, with `kubectl`, then we get the following output:
```shell
kubectl get clusteradmissionpolicies.policies.kubewarden.io
NAME               POLICY SERVER   MUTATING   STATUS
generated-policy   default         false      active
```

[Desired state]
It would be nice to have a "command switch" to set the name and if no value is provided, then the title of the policy is set (note: I used title to be aligned with `kwctl inspect` output)

Hope it makes sense 😄

ereslibre:
Thank you @nunix! I think it makes a lot of sense, and a nice UX improvement for kwctl.

PR #104, opened 2022-01-18
Add title to a policy created with kwctl manifest.

raulcabello:
fixes https://github.com/kubewarden/kwctl/issues/96

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -1,10 +1,13 @@
 use anyhow::{anyhow, Result};
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::convert::TryFrom;
 use validator::Validate;
 
 use policy_evaluator::policy_metadata::{Metadata, Rule};
+
+const POLICY_TITLE_ANNOTATION: &str = "io.kubewarden.policy.title";
 
 #[derive(Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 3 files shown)


src/manifest.rs
```diff
@@ -1,11 +1,14 @@
 use anyhow::{anyhow, Result};
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::convert::TryFrom;
 use validator::Validate;
 
 use policy_evaluator::policy_metadata::{Metadata, Rule};
 
+const POLICY_TITLE_ANNOTATION: &str = "io.kubewarden.policy.title";
```
ereslibre, line 10: We should use https://github.com/kubewarden/policy-evaluator/blob/ff05e8aa55fbe69f0a4a2bc0dc638c47a9cc9ec1/src/constants.rs#L2

ereslibre (changes requested):
Looking great, a comment to reuse a constant

src/manifest.rs
```diff
@@ -83,3 +92,80 @@ pub(crate) fn manifest(uri: &str, resource_type: &str, settings: Option<String>)
...
+            .map(|s| s.to_string())
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn mock_metadata_with_no_annotations() -> Metadata {
+        Metadata {
```
ereslibre, line 115: To some extent it's a pity that we don't implement `Default` for `Metadata`. It would make this tests much more concise. However, conceptually it probably doesn't make sense to derive it just for the sake of making usage easier. So yeah, nothing to do here :)

ereslibre (approved):
Thanks @raulcabello!

PR #169, opened 2022-02-25
feat: Keyless verification for kwctl {verify,pull,run} 

viccuad:
## Description

Part of https://github.com/kubewarden/policy-server/issues/142

Implement Sigstore keyless verification for`kwctl {verify,pull,run}`.
Now, accept new `--verification-config-path` that contains the config from [the Sigstore verification RFC](https://github.com/kubewarden/kubewarden-controller/blob/main/rfc/0003-policy-signing.md).
Previous flags are still supported: they get translated into a VerificationSettings config, to be used with the verification.

## Test

Run `make test e2e-test`.
Added e2e-tests that exercise `--verification-config-path`. 

## Additional Information

Depends on https://github.com/kubewarden/policy-fetcher/pull/55.

### Potential improvement

Create verification-config:
Provide better info on missing signatures: opened https://github.com/kubewarden/policy-fetcher/issues/57
Provide default verification-config: opened https://github.com/kubewarden/kwctl/issues/170

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ kube = { version = "0.68.0", default-features = false, features = ["client", "ru
 lazy_static = "1.4.0"
 mdcat = "0.26.1"
 policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.2.12" }
-policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.4.3" }
+policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", branch = "main" }
 pretty-bytes = "0.2.2"
 prettytable-rs = "^0.8"
 pulldown-cmark = { version = "0.9.1", default-features = false }
--- a/e2e-tests/e2e.bats
+++ b/e2e-tests/e2e.bats
@@ -83,8 +83,8 @@ kwctl() {
 
 @test "verify a signed policy from an OCI registry" {
     kwctl verify -a env=prod -a stable=true registry://ghcr.io/kubewarden/tests/pod-privileged:v0.1.9
-    [ "$status" -eq 2 ]
-    [ $(expr "$output" : '.*The following required arguments were not provided:.*') -ne 0 ]
+    [ "$status" -eq 1 ]
+    [ $(expr "$output" : '.*Intending to verify annotations, but no verification keys were passed.*') -ne 0 ]
     kwctl verify -k test-data/sigstore/cosign1.pub -k unexistent-path-to-key.pub -a env=prod -a stable=true registry://ghcr.io/kubewarden/tests/pod-privileged:v0.1.9
     [ "$status" -eq 1 ]
     [ $(expr "$output" : '.*No such file or directory.*') -ne 0 ]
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -21,7 +21,7 @@ kube = { version = "0.68.0", default-features = false, features = ["client", "ru
 lazy_static = "1.4.0"
 mdcat = "0.25.1"
 policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.2.12" }
-policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.4.3" }
```
viccuad, line 24: To be updated when https://github.com/kubewarden/policy-fetcher/pull/55 is in.

viccuad:
I'm puzzled by the GHA errors, GH is trying to pull a commit of policy-fetcher that doesn't exist anymore. It seems like a cache problem, but there's no way to clean the GHA cache manually.

src/main.rs
```diff
@@ -422,40 +393,85 @@ fn remote_server_options(matches: &ArgMatches) -> Result<(Option<Sources>, Optio
     Ok((sources, docker_config))
 }
 
-fn verification_options(
+fn verification_options(matches: &ArgMatches) -> Result<Option<VerificationSettings>> {
+    let verification_settings = if let Some(verification_config_path) =
```
flavio, line 397: I think we should error out if the user is specifying the configuration time and, at the same time, also pub keys and annotations.

src/main.rs
```diff
@@ -422,40 +393,85 @@ fn remote_server_options(matches: &ArgMatches) -> Result<(Option<Sources>, Optio
...
+        Some(read_verification_file(Path::new(
+            &verification_config_path,
+        ))?)
+    } else {
+        let verification_config_path = DEFAULT_ROOT.config_dir().join("verification-config.yaml");
+        if Path::exists(&verification_config_path) {
+            Some(read_verification_file(&verification_config_path)?)
+        } else {
+            build_verification_settings_from_flags(matches)?
+        }
```
flavio, line 409: I would give higher precedence to the cli flags provided by the user. Meaning: if the user provides pub keys/annotations constraints the configuration file on the disk is not going to be read.

e2e-tests/e2e.bats
```diff
@@ -96,11 +96,23 @@ kwctl() {
     [ $(expr "$output" : '.*Policy successfully verified.*') -ne 0 ]
     kwctl verify -k test-data/sigstore/cosign3.pub registry://ghcr.io/kubewarden/tests/pod-privileged:v0.1.9
     [ "$status" -eq 1 ]
-    [ $(expr "$output" : '.*No signing keys matched given constraints*') -ne 0 ]
+    [ $(expr "$output" : '.*Image verification failed: allOf list: No Signature Layer passed verification.*') -ne 0 ]
+    # TODO this shouldn't mention allOf, user has no notion of it
```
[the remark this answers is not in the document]
viccuad: In this case, it's doing `kwctl verify -k test-data/sigstore/cosign3.pub`. And then we internally create a VerificationSettings that we pass to policy-fetcher.
I think the user would be better served getting a list on which constraints are not satisfied, instead of the `allOf` message.

src/main.rs
viccuad: True! will do.

src/main.rs
viccuad: True! will do.

e2e-tests/e2e.bats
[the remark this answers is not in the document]
viccuad: Solved by https://github.com/sigstore/sigstore-rs/pull/42.

src/main.rs
viccuad: This can be done with:
```rust
    let verification_config = if let Some(verification_config_path) =
        matches.value_of("verification-config-path")
    {
        // verification-config has precedence over flags:
        if matches.is_present("verification-key") || matches.is_present("verification-annotation")
        {
            return Err(anyhow!(
                "verification-config-path flag has precedence over other verification flags"
            ));
        }
        Some(read_verification_file(Path::new(
            &verification_config_path,
        ))?)

```

But this is lossy. We will add new flags, and forget to add them to the list of flags that get checked for returning the error.

I would prefer to silently ignore flags, and just document that the config has precedence.

viccuad: From your next comment, giving higher precedence to cli flags from user instead of config, solves this.

src/main.rs
```diff
@@ -422,40 +393,85 @@ fn remote_server_options(matches: &ArgMatches) -> Result<(Option<Sources>, Optio
...
+    };
+    Ok(verification_settings)
+}
+
+// Takes clap flags and builds a Some(VerificationSettings) containing all
+// passed pub keys and annotations in VerificationSettings.AllOf.
+// If no verification flags where used, it returns a None.
+fn build_verification_settings_from_flags(
```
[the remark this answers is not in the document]
viccuad: Opened https://github.com/kubewarden/kwctl/issues/175.

viccuad:
Rebased on top of main and fixed conflicts, ready for review. Still pointing to my fork of policy-fetcher instead of a release.

flavio (approved):
LGTM, thanks

viccuad:
we need to release a new policy-fetcher and consume it in this pr. We either release policy-fetcher as it is, or we get https://github.com/kubewarden/policy-fetcher/issues/57 in, too.

viccuad:
Merging and consuming policy-fetcher from main, as we foresee that we will have several policy-fetcher releases shortly.

Issue #170, opened 2022-02-28
Feature Request: `kwctl scaffold-verification-config` command


viccuad:
### Is your feature request related to a problem?

Right now, users need to craft themselves a `verication-config.yml` for [the Sigstore verification RFC](https://github.com/kubewarden/kubewarden-controller/blob/main/rfc/0003-policy-signing.md#verifying-policies).

It would be nice to provide that config for them.

### Solution you'd like

A `kwctl scaffold-verification-config` or similar, that prints to stdout a valid verification config yaml. 
The printed config can be used as the default verification config of `kwctl` if piped into the default config path, effectively opting-in users into verifying policies by Sigstore from then onwards.

The config would be one that verifies kubewarden images.

The `kwctl scaffold-verification-config` could have subcommands for different types of configs:
1. `kwctl scaffold-verification-config permissive-kubewarden`: Verifies all kubewarden images (but only prints warnings on failed verifications, kubewarden or otherwise). Eg:
   ```yaml
   ---
   apiVersion: v1
   anyOf:
     minimumMatches: 0
     signatures:
       - kind: githubAction
         owner: kubewarden
   ```
1. `kwctl scaffold-verification-config enforce-kubewarden`: Verifies all kubewarden images, error if missing signatures
   ```yaml
   ---
   apiVersion: v1
   allOf:
   - kind: githubAction
     owner: kubewarden
   ```

We could also consider prepending a short commented paragraph explaining the config usage, and also commented examples as part of the command output (so the users can edit it directly without docs) when constructing the verification config.

### Alternatives you've considered

We have thought about transparently generating the config file by default if not present; yet users would miss the context on why they get rejected policies.

PR #183, opened 2022-03-16
feat: Add `kwctl scaffold manifest` & `kwctl scaffold verification-config`

viccuad:
## Description

Fix https://github.com/kubewarden/kwctl/issues/170

- Create mew `kwctl scaffold`:
  - Move `kwctl manifest` to `kwctl scaffold manifest` as subcommand.
  - Add `kwctl scaffold verification-config`:  Output default Sigstore verification configuration.

Example config:
```yaml
# Default Kubewarden verification config
#
# With this config, the only valid policies are those signed by Kubewarden
# infrastructure.
#
# This config can be saved to its default location (for this OS) with:
#   kwctl scaffold verification-config > /home/vic/.config/kubewarden/verification-config.yml
#
# Providing a config in the default location enables Sigstore verification.
# See https://docs.kubewarden.io for more Sigstore verification options.
---
apiVersion: v1
allOf:
  - kind: githubAction
    owner: kubewarden
    repo: ~
    annotations: ~
anyOf: ~
```

## Test
Added to `make e2e-test`

## Additional Information

Make bats e2e-tests run against tmp path.


Cargo.toml
```diff
@@ -5,6 +5,7 @@ version = "0.2.5"
 authors = [
         "Flavio Castelli <fcastelli@suse.com>",
         "Rafael Fernández López <rfernandezlopez@suse.com>",
+        "Víctor Cuadrado Juan <vcuadradojuan@suse.com>",
```
viccuad, line 8: This could be `Kubewarden Developers <kubewarden@suse.de>`, as once we add the missing, the printed string looks long.

ereslibre:
I would prefer to not add a command such as `create` to circumvent a limitation on the e2e test program.

Could other versions of `bats` work fine here as documented in https://bats-core.readthedocs.io/en/stable/gotchas.html#my-piped-command-does-not-work-under-run? (by doing `run bash -c 'echo foo | grep bar'`)

I would prefer to just have `kwctl scaffold-config`, maybe with relevant arguments, but that always prints to stdout.

I would also not make `scaffold` a subcommand of `kwctl verify`, seems like different things IMO. Writing the configuration to a different location is a matter of redirecting the output. If the output is a directory or can become a directory then `kwctl scaffold -d` seems fine IMO.

Cargo.toml
ereslibre: No objections, I'm fine replacing that.

e2e-tests/secure_supply_chain_tests.bats
```diff
@@ -1,10 +1,8 @@
 #!/usr/bin/env bats
 
-FULCIO_AND_REKORD_DATA_DIR=$(readlink -f ~/.config/kubewarden/fulcio_and_rekor_data)
-
 setup() {
     rm -rf ~/.cache/kubewarden
-    rm -rf ${FULCIO_AND_REKORD_DATA_DIR}
+    rm -rf ~/.config/kubewarden
```
jvanz, line 5: Cannot we mess up with our env using the home directory? Using a variable (even if the default value is the home directory), we can make the test portable to run in multiple directory each one with different configs. Maybe I'm being too pedantic here. But I'm just sharing a thought. ;-)

jvanz:
> I would prefer to not add a command such as `create` to circumvent a limitation on the e2e test program.

I also would say that if the e2e testing framework is a stone in our shoes, we should look for another options. Maybe we can use Ginko  to call commands and check the expected outputs. Another options is using pure bash, but then we will rewrite bat similar functions which seem to me reinvent the wheel.

e2e-tests/secure_supply_chain_tests.bats
viccuad: That's a good point, done in https://github.com/kubewarden/kwctl/pull/183/commits/13b95ae7fad71e232ad3460a152fdd1ffb743a5d. 
We could depend on bats-file and their [temp_file](https://github.com/ztombol/bats-file#working-with-temporary-directories), but I don't think it's worth the effort.

src/cli.rs
```diff
@@ -17,6 +19,24 @@ lazy_static! {
...
+            .unwrap_or_else(|| panic!("not possible to build base dirs"))
+            .home_dir()
+            .join(".sigstore")
+            .join("root")
+            .join("targets")
+    };
+    pub(crate) static ref SIGSTORE_FULCIO_CERT_PATH: PathBuf =
+        SIGSTORE_TARGETS_PATH.join("fulcio.crt.pem");
+    pub(crate) static ref SIGSTORE_REKOR_PUBLIC_KEY_PATH: PathBuf =
+        SIGSTORE_TARGETS_PATH.join("rekor.pub");
```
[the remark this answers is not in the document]
viccuad: Doesn't seem to, dropped in https://github.com/kubewarden/kwctl/pull/183/commits/d2b9f8b759f2aa509ef0f6f9595f987ec4554669.

Cargo.toml
viccuad: Done in https://github.com/kubewarden/kwctl/pull/183/commits/90c695d0f3690a0d9f63d2fc7a3252609e00c949.

viccuad:
Reimplemented, now the behaviour is: `kwctl scaffold-config verification`, with verification being a subcommand. There's no command for saving to file.

Thanks to @ereslibre, for noticing that we have [an alias for kwctl](https://github.com/viccuad/kwctl/blob/scaffold-config/e2e-tests/secure_supply_chain_tests.bats#L25) that has its own `run` inside. That was the culprit for losing the stdout and the docs workaround not applying :facepalm: . Solved now! :muscle:

viccuad:
Ready for review. Reimplemented as:
`kwctl scaffold` command, with
`kwctl scaffold verification-config` and
`kwctl scaffold manifest`, with old kwctl manifest flags

Updated PR description.

e2e-tests/secure_supply_chain_tests.bats
```diff
@@ -1,10 +1,24 @@
 #!/usr/bin/env bats
 
-FULCIO_AND_REKORD_DATA_DIR=$(readlink -f ~/.config/kubewarden/fulcio_and_rekor_data)
+xdg() {
```
ereslibre, line 3: Suggestion: rename to a more descriptive name: `setup_xdg_environment` or something along the lines.

ereslibre (approved):
Thanks @viccuad, minor comment

e2e-tests/secure_supply_chain_tests.bats
viccuad: done in last commit, https://github.com/kubewarden/kwctl/pull/183/commits/775dc214a8cd5bf5e847c9700316583976532a6f

ereslibre (approved):
Thanks @viccuad! 👏

flavio (approved):
LGTM

PR #190, opened 2022-03-25
feat: show Sigstore signatures on kwctl inspect

### LINKED ISSUE #181: Feature Request: show Sigstore signatures on `kwctl inspect`

As a user, I would like to obtain the list of signatures of an image, without the need for verifying them. This would be analogous to:
`crane manifest $(cosign triangulate ghcr.io/kubewarden/tests/pod-privileged:v0.1.9) | jq .`, and then inspecting the layer metadata by hand (see: https://github.com/sigstore/sigstore-rs/wiki/Key-based-signing-using-cosign-and-Rekor#the-signature-object).

### Solution you'd like

Enhance `kwctl inspect` to print information on all the Sigstore signatures (SignatureLayers) for a policy tag.

Note: Right now, `kwctl inspect` works on downloaded wasm policies in the store. This may need to change, as we would need to do a pull instead.

_No response_

### Anything else?

_No response_

raulcabello:
Display Sigstore signatures if available. Do nothing if they are not available.
Generate signatures url simulating the cosign triangulate output. For that we use the manifest digest, which returns the sha256 of the manifest content

Fix #181 

Output example:

```
Details              
title:              psp-capabilities 
description:        Replacement for the Kubernetes Pod Security Policy that controls the usage of capabilities 
author:             Flavio Castelli 
url:                https://github.com/kubewarden/capabilities-psp-policy 
source:             https://github.com/kubewarden/capabilities-psp-policy 
license:            Apache-2.0 
mutating:           true 
context aware:      false 
execution mode:     kubewarden-wapc 
protocol version:   1 
                     
Annotations          
io.kubewarden.kwctl 0.2.5-rc2 

Rules 
────────────────────
---
- apiGroups:
    - ""
  apiVersions:
    - v1
  resources:
    - pods
  operations:
    - CREATE
────────────────────

Usage 
This policy works by defining several lists of capabilities. Capabilities are specified as named in ALL_CAPS, but without the CAP_ prefix.

The following setting keys are accepted for this policy:

• allowed_capabilities: provides a list of capabilities that are allowed to be added to a container. The default set of capabilities are implicitly allowed. The empty set means that no additional capabilities may be added beyond the default set. * can be used to allow all capabilities.
• required_drop_capabilities: the capabilities which must be dropped from containers. These capabilities are removed from the default set, and must not be added. Capabilities listed in required_drop_capabilities must not be included in allowed_capabilities or default_add_capabilities.
• default_add_capabilities: the capabilities which are added to containers by default, in addition to the runtime defaults. See the documentation of your Container Runtime for the default list of capabilities.

The policy validates Pods at creation time and can also mutate them when either the required_drop_capabilities or the default_add_capabilities values are specified.

Sigstore signatures

Digest:                            sha256:081ed2cc2aa6ab05a4431d435e3d5eb8ea31ca2c56b6404384d124c2c0acdc71 
Media type:                        application/vnd.dev.cosign.simplesigning.v1+json 
Size:                              260 
Annotations                         
dev.cosignproject.cosign/signature MEUCIBNICT7dkhVCGwvrNN9FVOEXPYLrYP4zTZHjjTtUk5YNAiEA5JLfKSR+NWQFeS+Wn17hYwy3L/iEHZhwGMKp3WUdIlo= 
dev.sigstore.cosign/certificate    -----BEGIN CERTIFICATE----- 
                                   MIIDSTCCAs+gAwIBAgIUANEyn1gbXAKQPaWRpPGSe3NE7mgwCgYIKoZIzj0EAwMw 
                                   KjEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MREwDwYDVQQDEwhzaWdzdG9yZTAeFw0y 
                                   MjAxMjcxNzQwMzlaFw0yMjAxMjcxNzUwMzhaMBMxETAPBgNVBAoTCHNpZ3N0b3Jl 
                                   MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE1jH2vniruQuKS5hGktxQu5/yUSgx 
                                   cY2Ck3XhMM2oQRq6dNzMa4XCa6DBhXy+ZLRSP24j2ayQygvX4qOPZLhFxaOCAegw 
                                   ggHkMA4GA1UdDwEB/wQEAwIHgDATBgNVHSUEDDAKBggrBgEFBQcDAzAMBgNVHRMB 
                                   Af8EAjAAMB0GA1UdDgQWBBQcvAJYUgz0fUH/W+Kf4lFFd3+DHjAfBgNVHSMEGDAW 
                                   gBRYwB5fkUWlZql6zJChkyLQKsXF+jB4BgNVHREEcTBvhm1odHRwczovL2dpdGh1 
                                   Yi5jb20va3ViZXdhcmRlbi9naXRodWItYWN0aW9ucy8uZ2l0aHViL3dvcmtmbG93 
                                   cy9yZXVzYWJsZS1yZWxlYXNlLXBvbGljeS1ydXN0LnltbEByZWZzL2hlYWRzL3Yx 
                                   MB4GCisGAQQBg78wAQYEEHJlZnMvdGFncy92MC4xLjkwEgYKKwYBBAGDvzABAgQE 
                                   cHVzaDAcBgorBgEEAYO/MAEEBA5SZWxlYXNlIHBvbGljeTA2BgorBgEEAYO/MAED 
                                   BChiMjlhZGI1YjY3OWViNDRhOWFkY2Q1Y2YzNjQwZTM5ZGVkMTkwYzYyMDkGCisG 
                                   AQQBg78wAQEEK2h0dHBzOi8vdG9rZW4uYWN0aW9ucy5naXRodWJ1c2VyY29udGVu 
                                   dC5jb20wMAYKKwYBBAGDvzABBQQia3ViZXdhcmRlbi9jYXBhYmlsaXRpZXMtcHNw 
                                   LXBvbGljeTAKBggqhkjOPQQDAwNoADBlAjEAlhnA7/aWyHdKD4qRftfghXqIXBIV 
                                   eh2MULjOMdFJfSRsry9wbyizB+myxuvdtd8LAjBayJmG3+1w8Ci8eV7EafEPW3r3 
                                   N9IIG0oIXRot6MY7UNSCyBpcZKPc360zYe8z57k= 
                                   -----END CERTIFICATE----- 
dev.sigstore.cosign/timestamp      {"signatures":[{"keyid":"b6710623a30c010738e64c5209d367df1c0a18cf90e6ab5292fb01680f83453d","sig":"3045022100b4cda580b371e32ce938d03676208b394942f6205eb8ebe8650ec5b366ac5cd7022038ce51f5e3de90849ff728c21179484135b286563f60c906f0db5d999fd676d7"}],"signed":{"_type":"timestamp","spec_version":"1.0","version":10,"expires":"2022-01-29T00:38:38Z","meta":{"snapshot.json":{"length":1657,"hashes":{"sha256":"c4d26ba0e5c0b142c26b9cb11caedac1f29134275190da8f2cb981f2d8a13236","sha512":"db3c7a61418fd050a6af04f9d3bb2b359785fd5b4b323516740c76b413df0deae0d81f4bab0abce7abe077584110c00e0f35eeee5155abb7be1b74975794a8fe"},"version":10}}}} 
dev.sigstore.cosign/bundle         {"SignedEntryTimestamp":"MEUCIQCTsSmCHAnPYzjENHAUceoUqqjzNYyamfmXfmtjwszsDwIgHxzHN5c476RJZfS5xy6k7e1NiwUvqkr2w+Kd9u6/f2c=","Payload":{"body":"eyJhcGlWZXJzaW9uIjoiMC4wLjEiLCJraW5kIjoiaGFzaGVkcmVrb3JkIiwic3BlYyI6eyJkYXRhIjp7Imhhc2giOnsiYWxnb3JpdGhtIjoic2hhMjU2IiwidmFsdWUiOiIwODFlZDJjYzJhYTZhYjA1YTQ0MzFkNDM1ZTNkNWViOGVhMzFjYTJjNTZiNjQwNDM4NGQxMjRjMmMwYWNkYzcxIn19LCJzaWduYXR1cmUiOnsiY29udGVudCI6Ik1FVUNJQk5JQ1Q3ZGtoVkNHd3ZyTk45RlZPRVhQWUxyWVA0elRaSGpqVHRVazVZTkFpRUE1SkxmS1NSK05XUUZlUytXbjE3aFl3eTNML2lFSFpod0dNS3AzV1VkSWxvPSIsInB1YmxpY0tleSI6eyJjb250ZW50IjoiTFMwdExTMUNSVWRKVGlCRFJWSlVTVVpKUTBGVVJTMHRMUzB0Q2sxSlNVUlRWRU5EUVhNclowRjNTVUpCWjBsVlFVNUZlVzR4WjJKWVFVdFJVR0ZYVW5CUVIxTmxNMDVGTjIxbmQwTm5XVWxMYjFwSmVtb3dSVUYzVFhjS1MycEZWazFDVFVkQk1WVkZRMmhOVFdNeWJHNWpNMUoyWTIxVmRWcEhWakpOVWtWM1JIZFpSRlpSVVVSRmQyaDZZVmRrZW1SSE9YbGFWRUZsUm5jd2VRcE5ha0Y0VFdwamVFNTZVWGROZW14aFJuY3dlVTFxUVhoTmFtTjRUbnBWZDAxNmFHRk5RazE0UlZSQlVFSm5UbFpDUVc5VVEwaE9jRm96VGpCaU0wcHNDazFHYTNkRmQxbElTMjlhU1hwcU1FTkJVVmxKUzI5YVNYcHFNRVJCVVdORVVXZEJSVEZxU0RKMmJtbHlkVkYxUzFNMWFFZHJkSGhSZFRVdmVWVlRaM2dLWTFreVEyc3pXR2hOVFRKdlVWSnhObVJPZWsxaE5GaERZVFpFUW1oWWVTdGFURkpUVURJMGFqSmhlVkY1WjNaWU5IRlBVRnBNYUVaNFlVOURRV1ZuZHdwblowaHJUVUUwUjBFeFZXUkVkMFZDTDNkUlJVRjNTVWhuUkVGVVFtZE9Wa2hUVlVWRVJFRkxRbWRuY2tKblJVWkNVV05FUVhwQlRVSm5UbFpJVWsxQ0NrRm1PRVZCYWtGQlRVSXdSMEV4VldSRVoxRlhRa0pSWTNaQlNsbFZaM293WmxWSUwxY3JTMlkwYkVaR1pETXJSRWhxUVdaQ1owNVdTRk5OUlVkRVFWY0taMEpTV1hkQ05XWnJWVmRzV25Gc05ucEtRMmhyZVV4UlMzTllSaXRxUWpSQ1owNVdTRkpGUldOVVFuWm9iVEZ2WkVoU2QyTjZiM1pNTW1Sd1pFZG9NUXBaYVRWcVlqSXdkbUV6Vm1sYVdHUm9ZMjFTYkdKcE9XNWhXRkp2WkZkSmRGbFhUakJoVnpsMVkzazRkVm95YkRCaFNGWnBURE5rZG1OdGRHMWlSemt6Q21ONU9YbGFXRlo2V1ZkS2MxcFRNWGxhVjNoc1dWaE9iRXhZUW5aaVIyeHFaVk14ZVdSWVRqQk1ibXgwWWtWQ2VWcFhXbnBNTW1oc1dWZFNla3d6V1hnS1RVSTBSME5wYzBkQlVWRkNaemM0ZDBGUldVVkZTRXBzV201TmRtUkhSbTVqZVRreVRVTTBlRXhxYTNkRloxbExTM2RaUWtKQlIwUjJla0ZDUVdkUlJRcGpTRlo2WVVSQlkwSm5iM0pDWjBWRlFWbFBMMDFCUlVWQ1FUVlRXbGQ0YkZsWVRteEpTRUoyWWtkc2FtVlVRVEpDWjI5eVFtZEZSVUZaVHk5TlFVVkVDa0pEYUdsTmFteG9Xa2RKTVZscVdUTlBWMVpwVGtSU2FFOVhSbXRaTWxFeFdUSlplazVxVVhkYVZFMDFXa2RXYTAxVWEzZFplbGw1VFVSclIwTnBjMGNLUVZGUlFtYzNPSGRCVVVWRlN6Sm9NR1JJUW5wUGFUaDJaRWM1Y2xwWE5IVlpWMDR3WVZjNWRXTjVOVzVoV0ZKdlpGZEtNV015Vm5sWk1qbDFaRWRXZFFwa1F6VnFZakl3ZDAxQldVdExkMWxDUWtGSFJIWjZRVUpDVVZGcFlUTldhVnBZWkdoamJWSnNZbWs1YWxsWVFtaFpiV3h6WVZoU2NGcFlUWFJqU0U1M0NreFlRblppUjJ4cVpWUkJTMEpuWjNGb2EycFBVRkZSUkVGM1RtOUJSRUpzUVdwRlFXeG9ia0UzTDJGWGVVaGtTMFEwY1ZKbWRHWm5hRmh4U1ZoQ1NWWUtaV2d5VFZWTWFrOU5aRVpLWmxOU2MzSjVPWGRpZVdsNlFpdHRlWGgxZG1SMFpEaE1RV3BDWVhsS2JVY3pLekYzT0VOcE9HVldOMFZoWmtWUVZ6TnlNd3BPT1VsSlJ6QnZTVmhTYjNRMlRWazNWVTVUUTNsQ2NHTmFTMUJqTXpZd2VsbGxPSG8xTjJzOUNpMHRMUzB0UlU1RUlFTkZVbFJKUmtsRFFWUkZMUzB0TFMwSyJ9fX19","integratedTime":1643305239,"logIndex":1181814,"logID":"c0d23d6ad406973f9559f3ba2d1ca01f84147d8ffc5b8445c224f98b9591801d"}} 
dev.sigstore.cosign/chain           
                                   -----BEGIN CERTIFICATE----- 
                                   MIIB9zCCAXygAwIBAgIUALZNAPFdxHPwjeDloDwyYChAO/4wCgYIKoZIzj0EAwMw 
                                   KjEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MREwDwYDVQQDEwhzaWdzdG9yZTAeFw0y 
                                   MTEwMDcxMzU2NTlaFw0zMTEwMDUxMzU2NThaMCoxFTATBgNVBAoTDHNpZ3N0b3Jl 
                                   LmRldjERMA8GA1UEAxMIc2lnc3RvcmUwdjAQBgcqhkjOPQIBBgUrgQQAIgNiAAT7 
                                   XeFT4rb3PQGwS4IajtLk3/OlnpgangaBclYpsYBr5i+4ynB07ceb3LP0OIOZdxex 
                                   X69c5iVuyJRQ+Hz05yi+UF3uBWAlHpiS5sh0+H2GHE7SXrk1EC5m1Tr19L9gg92j 
                                   YzBhMA4GA1UdDwEB/wQEAwIBBjAPBgNVHRMBAf8EBTADAQH/MB0GA1UdDgQWBBRY 
                                   wB5fkUWlZql6zJChkyLQKsXF+jAfBgNVHSMEGDAWgBRYwB5fkUWlZql6zJChkyLQ 
                                   KsXF+jAKBggqhkjOPQQDAwNpADBmAjEAj1nHeXZp+13NWBNa+EDsDP8G1WWg1tCM 
                                   WP/WHPqpaVo0jhsweNFZgSs0eE7wYI4qAjEA2WB9ot98sIkoF3vZYdd3/VtWB5b9 
                                   TNMea7Ix/stJ5TfcLLeABLE4BNJOsQ4vnBHJ 
                                   -----END CERTIFICATE----- 
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ k8s-openapi = { version = "0.14.0", default-features = false, features = ["v1_22
 kube = { version = "0.68.0", default-features = false, features = ["client", "rustls-tls"] }
 lazy_static = "1.4.0"
 mdcat = "0.26.1"
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.2.16" }
+policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.2.17" }
 pretty-bytes = "0.2.2"
 prettytable-rs = "^0.8"
 pulldown-cmark = { version = "0.9.1", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,6 @@ url = "2.2.2"
 validator = { version = "0.14", features = ["derive"] }
 walrus = "0.19.0"
 wasmparser = "0.83.0"
-oci-distribution = { git = "https://github.com/krustlet/oci-distribution", rev = "0f717968093a5415f428503d741dedf24ea97948", default-features = false, features =  ["rustls-tls"] }
 
 [dev-dependencies]
 tempfile = "3.3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,8 @@ url = "2.2.2"
 validator = { version = "0.14", features = ["derive"] }
 walrus = "0.19.0"
 wasmparser = "0.83.0"
+oci-distribution = { git = "https://github.com/krustlet/oci-distribution", rev = "0f717968093a5415f428503d741dedf24ea97948", default-features = false, features =  ["rustls-tls"] }
 
 [dev-dependencies]
 tempfile = "3.3.0"
+rstest = "0.12.0"
--- a/src/inspect.rs
+++ b/src/inspect.rs
@@ -20,7 +20,7 @@ pub(crate) async fn inspect(
 ) -> Result<()> {
     let uri = crate::utils::map_path_to_uri(uri)?;
     let wasm_path = crate::utils::wasm_path(uri.as_str())?;
-    let printer = get_printer(&output);
+    let metadata_printer = MetadataPrinter::from(&output);
 
     let metadata = Metadata::from_path(&wasm_path)
         .map_err(|e| anyhow!("Error parsing policy metadata: {}", e))?;
@@ -28,7 +28,7 @@ pub(crate) async fn inspect(
     let signatures = fetch_signatures_manifest(uri.as_str(), sources, docker_config).await;
 
     match metadata {
-        Some(metadata) => printer.print(&metadata)?,
+        Some(metadata) => metadata_printer.print(&metadata)?,
         None => return Err(anyhow!(
             "No Kubewarden metadata found inside of '{}'.\nPolicies can be annotated with the `kwctl annotate` command.",
             uri
@@ -41,7 +41,7 @@ pub(crate) async fn inspect(
                 println!();
                 println!("Sigstore signatures");
                 println!();
-                let sigstore_printer = get_signatures_printer(output);
+                let sigstore_printer = SignaturesPrinter::from(&output);
                 sigstore_printer.print(&signatures);
             }
         }
@@ -79,30 +79,38 @@ impl TryFrom<Option<&str>> for OutputType {
     }
 }
 
-fn get_printer(output_type: &OutputType) -> Box<dyn MetadataPrinter> {
-    match output_type {
-        OutputType::Yaml => Box::new(MetadataYamlPrinter {}),
-        OutputType::Pretty => Box::new(MetadataPrettyPrinter {}),
+enum MetadataPrinter {
+    Yaml,
+    Pretty,
+}
+
+impl From<&OutputType> for MetadataPrinter {
+    fn from(output_type: &OutputType) -> Self {
+        match output_type {
+            OutputType::Yaml => Self::Yaml,
+            OutputType::Pretty => Self::Pretty
+        }
     }
 }
 
-trait MetadataPrinter {
-    fn print(&self, metadata: &Metadata) -> Result<()>;
-}
-
-struct MetadataYamlPrinter {}
-
-impl MetadataPrinter for MetadataYamlPrinter {
-    fn print(&self, metadata: &Metadata) -> Result<()> {
-        let metadata_yaml = serde_yaml::to_string(&metadata)?;
-        println!("{}", metadata_yaml);
-        Ok(())
+impl MetadataPrinter {
+    fn print(&self, metadata: &Metadata) -> Result<()>{
+        match self {
+            MetadataPrinter::Yaml => {
+                let metadata_yaml = serde_yaml::to_string(metadata)?;
+                println!("{}", metadata_yaml);
+                Ok(())
+            }
+            MetadataPrinter::Pretty => {
+                self.print_metadata_generic_info(metadata)?;
+                println!();
+                self.print_metadata_rules(metadata)?;
+                println!();
+                self.print_metadata_usage(metadata)
+            }
+        }
     }
-}
 
-struct MetadataPrettyPrinter {}
-
-impl MetadataPrettyPrinter {
     fn annotation_to_row_key(&self, text: &str) -> String {
         let mut out = String::from(text);
         out.push(':');
@@ -218,56 +226,46 @@ impl MetadataPrettyPrinter {
     }
 }
 
-impl MetadataPrinter for MetadataPrettyPrinter {
-    fn print(&self, metadata: &Metadata) -> Result<()> {
-        self.print_metadata_generic_info(metadata)?;
-        println!();
-        self.print_metadata_rules(metadata)?;
-        println!();
-        self.print_metadata_usage(metadata)
-    }
+enum SignaturesPrinter {
+    Yaml,
+    Pretty,
 }
 
-fn get_signatures_printer(output_type: OutputType) -> Box<dyn SignaturesPrinter> {
-    match output_type {
-        OutputType::Yaml => Box::new(SignaturesYamlPrinter {}),
-        OutputType::Pretty => Box::new(SignaturesPrettyPrinter {}),
-    }
-}
-
-trait SignaturesPrinter {
-    fn print(&self, signatures: &OciImageManifest);
-}
-
-struct SignaturesPrettyPrinter {}
-
-impl SignaturesPrinter for SignaturesPrettyPrinter {
-    fn print(&self, signatures: &OciImageManifest) {
-        for layer in &signatures.layers {
-            let mut table = Table::new();
-            table.set_format(FormatBuilder::new().padding(0, 1).build());
-            table.add_row(row![Fmbl -> "Digest: ", layer.digest]);
-            table.add_row(row![Fmbl -> "Media type: ", layer.media_type]);
-            table.add_row(row![Fmbl -> "Size: ", layer.size]);
-            if let Some(annotations) = &layer.annotations {
-                table.add_row(row![Fmbl -> "Annotations"]);
-                for annotation in annotations.iter() {
-                    table.add_row(row![Fgbl -> annotation.0, annotation.1]);
-                }
-            }
-            table.printstd();
-            println!();
+impl From<&OutputType> for SignaturesPrinter {
+    fn from(output_type: &OutputType) -> Self {
+        match output_type {
+            OutputType::Yaml => Self::Yaml,
+            OutputType::Pretty => Self::Pretty
         }
     }
 }
 
-struct SignaturesYamlPrinter {}
-
-impl SignaturesPrinter for SignaturesYamlPrinter {
+impl SignaturesPrinter {
     fn print(&self, signatures: &OciImageManifest) {
-        let signatures_yaml = serde_yaml::to_string(signatures);
-        if let Ok(signatures_yaml) = signatures_yaml {
-            println!("{}", signatures_yaml)
+        match self {
+            SignaturesPrinter::Yaml => {
+                let signatures_yaml = serde_yaml::to_string(signatures);
+                if let Ok(signatures_yaml) = signatures_yaml {
+                    println!("{}", signatures_yaml)
+                }
+            }
+            SignaturesPrinter::Pretty => {
+                    for layer in &signatures.layers {
+                        let mut table = Table::new();
+                        table.set_format(FormatBuilder::new().padding(0, 1).build());
+                        table.add_row(row![Fmbl -> "Digest: ", layer.digest]);
+                        table.add_row(row![Fmbl -> "Media type: ", layer.media_type]);
+                        table.add_row(row![Fmbl -> "Size: ", layer.size]);
+                        if let Some(annotations) = &layer.annotations {
+                            table.add_row(row![Fmbl -> "Annotations"]);
+                            for annotation in annotations.iter() {
+                                table.add_row(row![Fgbl -> annotation.0, annotation.1]);
+                            }
+                        }
+                        table.printstd();
+                        println!();
+                    }
+            }
         }
     }
 }
--- a/src/inspect.rs
+++ b/src/inspect.rs
@@ -1,7 +1,7 @@
 use crate::{DockerConfig, Registry, Sources};
 use anyhow::{anyhow, Result};
 use mdcat::{ResourceAccess, TerminalCapabilities, TerminalSize};
-use oci_distribution::manifest::{OciImageManifest, OciManifest};
+use policy_evaluator::policy_fetcher::oci_distribution::manifest::{OciImageManifest, OciManifest};
 use policy_evaluator::policy_fetcher::sigstore::cosign::{ClientBuilder, CosignCapabilities};
 use policy_evaluator::policy_fetcher::sigstore::registry::{Auth, ClientConfig};
 use policy_evaluator::{
--- a/src/inspect.rs
+++ b/src/inspect.rs
@@ -2,12 +2,14 @@ use crate::{DockerConfig, Registry, Sources};
 use anyhow::{anyhow, Result};
 use mdcat::{ResourceAccess, TerminalCapabilities, TerminalSize};
 use oci_distribution::manifest::{OciImageManifest, OciManifest};
+use policy_evaluator::policy_fetcher::sigstore::cosign::{ClientBuilder, CosignCapabilities};
+use policy_evaluator::policy_fetcher::sigstore::registry::{Auth, ClientConfig};
 use policy_evaluator::{
     constants::*, policy_evaluator::PolicyExecutionMode, policy_metadata::Metadata,
 };
 use prettytable::{format::FormatBuilder, Table};
 use pulldown_cmark::{Options, Parser};
-use std::convert::TryFrom;
+use std::convert::{TryFrom, TryInto};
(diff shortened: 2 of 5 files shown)


jvanz (approved):
LGTM! :D

src/inspect.rs
```diff
@@ -8,20 +10,38 @@ use pulldown_cmark::{Options, Parser};
...
+) -> Result<()> {
     let uri = crate::utils::map_path_to_uri(uri)?;
     let wasm_path = crate::utils::wasm_path(uri.as_str())?;
-    let printer = get_printer(output);
+    let printer = get_printer(&output);
 
     let metadata = Metadata::from_path(&wasm_path)
         .map_err(|e| anyhow!("Error parsing policy metadata: {}", e))?;
+
+    let signatures = fetch_signatures_manifest(uri.as_str(), sources, docker_config).await;
```
flavio, line 26: I would move this line below the `match metadata` block.

Cargo.toml
```diff
@@ -36,6 +36,8 @@ url = "2.2.2"
 validator = { version = "0.14", features = ["derive"] }
 walrus = "0.19.0"
 wasmparser = "0.83.0"
+oci-distribution = { git = "https://github.com/krustlet/oci-distribution", rev = "0f717968093a5415f428503d741dedf24ea97948", default-features = false, features =  ["rustls-tls"] }
```
flavio, line 39: AFAIK this is being exposed by `policy-fetcher`. Try to consume it through the export done by `policy-fetcher`

src/inspect.rs
```diff
@@ -189,3 +209,106 @@ impl MetadataPrinter for MetadataPrettyPrinter {
         self.print_metadata_usage(metadata)
     }
 }
+
+fn get_signatures_printer(output_type: OutputType) -> Box<dyn SignaturesPrinter> {
```
flavio, line 213: While using the trait system is correct, I feel like this is kinda an overkill in this scenario and I'm not so sure about that being idiomatic.

I think this could be solved by defining `SignaturePrinter` to be an `enum` with two values: `Yaml` and `Pretty`. Then the `print` method would behave differently based on the type of the enum.

@ereslibre what do you think? I'm interested to hear your opinion about the idiomatic part. I do not intend to nitpick on this code written by @raulcabello :)

src/inspect.rs
```diff
@@ -189,3 +209,106 @@ impl MetadataPrinter for MetadataPrettyPrinter {
...
+            println!("{}", signatures_yaml)
+        }
+    }
+}
+
+async fn fetch_signatures_manifest(
+    uri: &str,
+    sources: Option<Sources>,
+    docker_config: Option<DockerConfig>,
+) -> Option<OciImageManifest> {
```
flavio, line 261: I think it would be better to change the return value of  this function to be `Result<Option<OciImageManifest>>`.

The current code swallows errors and turn them into `None`. That can lead the user to assume a policy is not signed, while instead there's just a network hiccup that prevents the code from fetching the signatures.

Note well: I don't think we should exit with an error as soon as the signatures cannot be fetched. We should still print all the information about the policy and then, inside of the signature section, we should write something like "Cannot determine if the policy has been signed. There was an error while attempting to fetch its signatures from the remote registry: <error>".

src/inspect.rs
```diff
@@ -189,3 +209,106 @@ impl MetadataPrinter for MetadataPrettyPrinter {
...
+        .ok()?;
+
+    match manifest {
+        OciManifest::Image(img) => Some(img),
+        _ => None,
+    }
+}
+
+// simulate cosign triangulate output. Use the manifest digest, which returns the sha256 of the manifest content, then replace ':' with '-' and append '.sig'
+fn get_signature_url(mut uri: String, digest: &str) -> Option<String> {
```
flavio, line 277: Why do we have to simulate the triangulate function of sigstore? I'm pretty sure `policy-fetcher` is already exposing sigstore, hence we can just use the original function.

If `policy-fetcher` is not exposing sigstore, we can change it

flavio (changes requested):
Overall looks good to me, but I left some notes about code that has to be changed

src/inspect.rs
ereslibre: Yes, I agree. Even something like `PrinterOutput`, `OutputFormat` or something similar, I think is closer to the intent of the parameter, but I think is better to not use traits in a similar way as we would use go interfaces, with Rust we have a more powerful type system :)

src/inspect.rs
raulcabello: If I move this line below the match, there will be a delay when displaying the signatures.
It will first print the metadata, then do a http request to fetch the signatures and display them. It feels a bit strange for the user to see the metada printed and after maybe one or two seconds (probably less depending on the network) the signatures. What do you think?

Cargo.toml
raulcabello: it is used in `policy-fetcher` but it is not publicly exposed. kwctl does not have a direct dependency of `policy-fetcher`. 
I can export oci-distribution in `policy-fetcher` and then in `policy-evaluator`. Would this be ok? or am I missing something?
if this is ok I'll submit PRs for `policy-fetcher` and `policy-evaluator`

src/inspect.rs
raulcabello: I'm still fairly new to rust and I'd love to write more idiomatic code. So this feedback is very appreciated! 
Should I also refactor the [MetadataPrinter](https://github.com/kubewarden/kwctl/blob/main/src/inspect.rs#L44-L65)?
Maybe it makes more sense to merge MetadataPrinter and SignaturesPrinter into one Printer?

ereslibre: I think having two printers is fine, given they are focused on printing different types of information.

For uniformity I would refactor the `MetadataPrinter` as well, yes.

raulcabello: done with enums https://github.com/kubewarden/kwctl/pull/190/commits/2a554bd961017abb09befb5ac28edba4dd82717d

src/inspect.rs
```diff
@@ -1,27 +1,65 @@
+use crate::{DockerConfig, Registry, Sources};
 use anyhow::{anyhow, Result};
 use mdcat::{ResourceAccess, TerminalCapabilities, TerminalSize};
+use policy_evaluator::policy_fetcher::oci_distribution::manifest::{OciImageManifest, OciManifest};
```
flavio, line 4: there are some clippy warnings that are concerning, like

```
failed to resolve: could not find `oci_distribution` in `policy_fetcher`
```

Maybe we have to rerun the tests?

src/inspect.rs
flavio: Makes sense, thanks for the explanation

flavio (approved):
LGTM, let's just make sure everything is building (see the comment about clippy errors)

flavio (approved):
LGTM :rocket:

src/inspect.rs
```diff
@@ -1,27 +1,65 @@
...
+use policy_evaluator::policy_fetcher::sigstore::cosign::{ClientBuilder, CosignCapabilities};
+use policy_evaluator::policy_fetcher::sigstore::registry::{Auth, ClientConfig};
```
ereslibre, line 6:
```suggestion
use policy_evaluator::policy_fetcher::{
  manifest::{OciImageManifest, OciManifest},
  sigstore::{
    cosign::{
      ClientBuilder,
      CosignCapabilities,
    },
    registry::{Auth, ClientConfig}
  },
};
```

ereslibre: Non-blocking comment, it's a bit less repetitive. Feel free to ignore or change spacing etc...

ereslibre (commented):
LGTM, thanks! A minor comment regarding the use statements.

PR #198, opened 2022-04-12
Update to latest version of wasmparser

flavio:
Update to latest version, plus fix the code to ensure it builds after the API has been changed.

Moreover, include some e2e tests that are using this portion of the code.


jvanz (approved):
LGTM! :D

PR #212, opened 2022-04-29
Support policies performing revers DNS lookups

flavio:
Update to latest version of policy-evaluator, so that new waPC functions are supported.

This is needed to fix https://github.com/kubewarden/policy-server/issues/236

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -267,7 +267,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "miniz_oxide",
- "object 0.28.3",
+ "object 0.28.4",
  "rustc-demangle",
 ]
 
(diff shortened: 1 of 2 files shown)


viccuad (approved):
LGTM, Straight bump (needs to consume correct dep in cargo.toml)

ereslibre (commented):
LGTM modulo the tiny "reverse" detail in the commit description. With that meaning: I think we should get rid of the "reverse" word given we are doing regular DNS resolutions.

flavio:
I've updated the `Cargo.toml` to reference the just released version of `policy-evaluator`. I'm going to merge it once all the tests are green.

BTW, I've also fixed the commit message as pointed out by @ereslibre

flavio:
GH got stuck during the e2e tests because of unknown reasons. All the tests are green locally, going ahead with the merge

Issue #240, opened 2022-06-15
Feature Request: suppress warning messages generated by Wasmtime Cache


flavio:
### Is your feature request related to a problem?

Sometimes, when wasmtime cannot update its own cache, a warning message is printed on the stderr.

This looks like that:

```console
[...]

PR #241, opened 2022-06-15
Suppress wasmtime cache warnings

flavio:
The warning messages are not critical, but they confuse the users.

Moreover, they pollute the stdout, leading to end to end test breakage.

Fixes https://github.com/kubewarden/kwctl/issues/240


viccuad (approved):
LGTM!

Issue #208, opened 2022-04-25
AWS ECR Authorization not recognized after Docker Login


Jawery:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When trying to interact with aws ECR to push or pull an artifact I'm getting a 401 unauthorized even after a successful `docker login` and display of ability to push/pull normal docker images. Verified a few things over on the slack channel so I'll repeat them here for simplicity:

- Logging in to docker from the cli piping the result from `aws ecr get-login-password` to the login command.
- Verified I was using a token that wasn't expired
- Tried passing in the .docker/config.json file itself with no change
- Same functionality on v0.2.5 and v0.3.0

### Expected Behavior

Authentication through docker cli to aws ecr enables kwctl interaction with aws ecr as well. Alternatively some other auth solution for aws would work as well.

### Steps To Reproduce

1. `Docker Login` to aws ecr
2. Verify that `Docker push/pull [some ecr repo]` is successful after authentication
4. `kwctl push [some ecr repo]`
5. Here I'm encountering a 401 response from aws

### Environment

```markdown
- OS: Mac
- Architecture:x86-64
```

viccuad:
As a workaround, it should be possible to use [`wasm-to-oci`](https://github.com/engineerd/wasm-to-oci), which lists aws ECR. An  example pushing to GH packages:
```
$ git clone git@github.com:kubewarden/pod-privileged-policy.git
$ cd pod-privileged-policy
$ make annotated-policy.wasm
$ wasm-to-oci push annotated-policy.wasm ghcr.io/viccuad/policies/pod-privileged:wasm-to-oci
INFO[0005] Pushed: ghcr.io/viccuad/policies/pod-privileged:wasm-to-oci
INFO[0005] Size: 21892
INFO[0005] Digest: sha256:4ada0a5ceda88ef69f26ab11213e99f5595465442ed3b5665a868ac3d919ef0e
```

Jawery:
As mentioned in the slack channel unfortunately it seems `wasm-to-oci` isn't working with AWS ECR currently, when the policyServer tries to pull an artifact uploaded using this tool from AWS ECR the logs display an error `policy could not be downloaded due to error: error decoding response body: expected value at line 1 column 1`

[...]

PR #277, opened 2022-08-09
feat: use docker_credential crate instead of DockerConfig

raulcabello:
fix https://github.com/kubewarden/kwctl/issues/208


src/inspect.rs
```diff
@@ -277,33 +273,18 @@ impl SignaturesPrinter {
...
-
-            match sigstore_auth {
-                None => Auth::Anonymous,
-                Some(sa) => sa?,
-            }
-        }
-        None => Auth::Anonymous,
+    let auth = match Registry::auth(image_name) {
+        RegistryAuth::Anonymous => Auth::Anonymous,
+        RegistryAuth::Basic(username, password) => Auth::Basic(username, password),
```
flavio, line 287: It would be great to have the final code look something like that:

```rust
let auth: Auth = Registry::auth(image_image).into()
```

src/main.rs
```diff
@@ -409,23 +395,11 @@ fn remote_server_options(matches: &ArgMatches) -> Result<(Option<Sources>, Optio
...
-            Some(read_docker_config_json_file(&config_json_path)?)
-        } else {
-            None
-        }
-    } else {
-        None
-    };
-    Ok((sources, docker_config))
+    if let Some(docker_config_json_path) = matches.get_one::<String>("docker-config-json-path") {
+        env::set_var(DOCKER_CONFIG_ENV_VAR, docker_config_json_path);
```
flavio, line 399: nit, add a comment explaining why we are messing around with the env variable

flavio (commented):
LGTM, I left some minor comments

src/main.rs
raulcabello: https://github.com/kubewarden/kwctl/pull/277/commits/b3c66376dba2c5b7bcde8ad396225b2fb384ea21

src/inspect.rs
flavio: nah, I think this is not doable. Let's leave the code as it currently is

flavio (approved):
ship it!

flavio:
@raulcabello merge it, but wait a little before tagging a new release. I think it would be great to have https://github.com/kubewarden/kwctl/pull/278 merged before the next patch release of kwctl. In that way we would have apple silicon binaries too!

PR #289, opened 2022-08-19
 Use Musl from Ubuntu packages.

### LINKED ISSUE #281: Building release is flacky
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

From time to time, building the releases of kwctl fails. The failure happens because there's an error downloading the musl release we use to perform the cross build.

This is really bad, because it doesn't allow us to release new features/fixes.

### Expected Behavior

First of all, the build should be stable. It should be green every time.

While addressing this issue, it would be nice (but not mandatory) to address also these current limitations:

    We are downloading a musl release that is not versioned. This makes things not reproducible and less predictable.
    When doing a SBOM, this would be problematic

Currently we are using a two stage build process. The 1st stage builds a statically linked binary for policy-server, which is then copied into a distroless container.

Maybe we can perform stage 1 inside some linux distribution container image which comes with the musl toolkit.

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- Architecture:
```

### Anything else?

_No response_

jvanz:
After a DDoS attack the maintainer of the musl.cc domains blocked the
request caming from some cloud providers. This prevent us from
downloading the tarball needed to cross-compile the policy-server.
Therefore, the workflow has been updated to use the musl package
available in the Ubuntu 22.04 packages.

Fix #281

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,75 +11,91 @@ jobs:
     # A branch is required, and cannot be dynamic - https://github.com/actions/runner/issues/1493
     uses: kubewarden/kwctl/.github/workflows/tests.yml@main
 
-  build-linux-x86_64:
-    name: Build linux (x86_64) binary
-    runs-on: ubuntu-latest
+  build-linux-binaries:
+    name: Build linux binary
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        targetarch: [ "aarch64", "x86_64" ]
+        include:
+          - targetarch: aarch64
+            arch: arm64
+            rustflags: ""
+          - targetarch: x86_64
+            arch: amd64
+            rustflags: "-C target-feature=+crt-static"
     permissions:
+      packages: write
       id-token: write
     needs:
       - ci
     steps:
-    - uses: actions/checkout@v2
-    - uses: sigstore/cosign-installer@main
-    - name: Setup rust toolchain
-      uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-    - name: Setup musl for x86_64
-      run: |
-        curl https://musl.cc/x86_64-linux-musl-cross.tgz | tar -xz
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,5 +1,7 @@
 name: kwctl release
 on:
+  # User can trigger this workflow in the Github UI
+  workflow_dispatch:
```
viccuad, line 4: for other workflows this is useful, but for the release one, I would prefer not to have it so releases are released once and never overwritten.

.github/workflows/release.yml
```diff
@@ -11,75 +13,77 @@ jobs:
...
+          sudo sed -i "$ a deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports jammy-security main restricted" /etc/apt/sources.list
+          sudo sed -i "$ a deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports jammy-security universe" /etc/apt/sources.list
+          sudo sed -i "$ a deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports jammy-security multiverse" /etc/apt/sources.list
+
+          sudo apt update -y
+      - name: Install Musl and configure gcc spec
+        run: |
+          sudo apt install -y musl-dev:${{ matrix.arch }}
+          sudo sed -i 1d /usr/lib/${{ matrix.targetarch }}-linux-musl/musl-gcc.specs
+          sudo sed -i "s/*cpp_options/+cpp_options/g" /usr/lib/${{ matrix.targetarch }}-linux-musl/musl-gcc.specs
```
viccuad, line 53: I would like to add a comment before the 2 lines patching the spec. Something like:
`# patching the .spec file, as by default it has a bug where it tries to set old_cpp_options but it already exists. using *+cpp_options achieves the same desired functionality of appending preexisting options`

viccuad (commented):
LGTM! just some comments

viccuad (approved):

PR #297, opened 2022-09-05
SBOM file generation.

### LINKED ISSUE #279: kwctl: create SBOM file
Each release should have a SBOM file that allows to track the origin of all the dependencies.

jvanz:
## Description

Adds Makefile target and update the release workflow to generate the SBOM file for the kwctl binaries.

Fix #279

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,6 @@ on:
 env:
   CARGO_TERM_COLOR: always
 jobs:
-
   ci:
     # A branch is required, and cannot be dynamic - https://github.com/actions/runner/issues/1493
     uses: kubewarden/kwctl/.github/workflows/tests.yml@main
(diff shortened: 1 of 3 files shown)


viccuad (approved):
LGTM!

(there's a bit of repetition between all build jobs, but that was before, and templating and using variables for all of that sounds supercumbersome)

raulcabello (approved):
LGTM

.github/workflows/release.yml
```diff
@@ -97,6 +96,30 @@ jobs:
           name: kwctl-linux-${{ matrix.targetarch }}
           path: kwctl-linux-${{ matrix.targetarch }}.zip
 
+      - name: Generate SBOM
+        run: |
+          make download-spdx-sbom-generator sbom
```
flavio, line 101: It's a matter of personal taste, but what about using https://github.com/kubewarden/github-actions/tree/v1/sbom-generator-installer to get the sbom tool? Just to be consistent with what we are doing with our policies

flavio (commented):
I know this got merged, I'm fine with that. I just have an extra question

.github/workflows/release.yml
jvanz: True. Let me fix it.

PR #299, opened 2022-09-07
Use github action to install SBOM generation tool.

jvanz:
Instead of download the SBOM generation tool in the workflow use the github action used in other repositories.

Depends on https://github.com/kubewarden/github-actions/pull/39


viccuad (approved):
LGTM. I'm ok removing the `make download-spdx-sbom-generator` target, we are gonna only need it on releases afterall.

raulcabello (approved):

PR #348, opened 2022-11-10
save and load commands

### LINKED ISSUE #346: kwctl save and load commands
Add a new `save` command that allows user to store policies in a tar file. Similar to the [docker save command](https://docs.docker.com/engine/reference/commandline/save/)

All policies provided will be stored in a tar file. Policies must already be available in the system (`kwctl pull` must be run before `save`)

```
kwctl save [OPTIONS] POLICY [POLICY...]

OPTIONS
--output , -o Write to a file, instead of STDOUT
```

Add a new 'load' commands that load policies inside a tarball into the default store

```
kwctl load --input <input>
```

raulcabello:
There are two new commands in this PR:

- `save`: stores the provided policies into a tarball
- `load`: load the policies in the tarball into the default store

This is needed for airgap support. Policies will be stored into a tarball, then this tarball is moved to the airgapped environment. Finally, `kwctl load` is used to import the policies in the airgapped environment.

Fix #346

### DIFF
--- a/e2e-tests/e2e.bats
+++ b/e2e-tests/e2e.bats
@@ -104,21 +104,21 @@ kwctl() {
 }
 
 @test "save and load" {
-    kwctl pull registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9
+    kwctl pull registry://ghcr.io/kubewarden/tests/pod-privileged:v0.1.9
     kwctl pull https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm
     kwctl policies
     [[ $(echo "$output" | wc -l) -eq 6 ]]
-    [[ "$output" =~ "registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9" ]]
+    [[ "$output" =~ "registry://ghcr.io/kubewarden/tests/pod-privileged:v0.1.9" ]]
     [[ "$output" =~ "https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm" ]]
-    kwctl save --output policies.tar.gz registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9 https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm
-    kwctl rm registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9
+    kwctl save --output policies.tar.gz registry://ghcr.io/kubewarden/tests/pod-privileged:v0.1.9 https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm
+    kwctl rm registry://ghcr.io/kubewarden/tests/pod-privileged:v0.1.9
     kwctl rm https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm
     kwctl policies
     [ "$output" = "" ]
     kwctl load --input policies.tar.gz
     kwctl policies
     [[ $(echo "$output" | wc -l) -eq 6 ]]
-    [[ "$output" =~ "registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9" ]]
+    [[ "$output" =~ "registry://ghcr.io/kubewarden/tests/pod-privileged:v0.1.9" ]]
     [[ "$output" =~ "https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm" ]]
     rm policies.tar.gz
 }
--- a/e2e-tests/e2e.bats
+++ b/e2e-tests/e2e.bats
@@ -102,3 +102,23 @@ kwctl() {
     [ "$status" -ne 0 ]
     [[ "$output" == "Error: Wrong value inside of policy's metatada for 'executionMode'. This policy has been created using Rego" ]]
 }
+
+@test "save and load" {
+    kwctl pull registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9
+    kwctl pull https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm
+    kwctl policies
+    [[ $(echo "$output" | wc -l) -eq 6 ]]
+    [[ "$output" =~ "registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9" ]]
+    [[ "$output" =~ "https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm" ]]
+    kwctl save --output policies.tar.gz registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9 https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm
+    kwctl rm registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9
+    kwctl rm https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm
+    kwctl policies
+    [ "$output" = "" ]
+    kwctl load --input policies.tar.gz
+    kwctl policies
+    [[ $(echo "$output" | wc -l) -eq 6 ]]
+    [[ "$output" =~ "registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9" ]]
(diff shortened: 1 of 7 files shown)


jvanz (approved):
LGTM! :D

src/cli.rs
```diff
@@ -527,6 +527,35 @@ pub fn build_cli() -> Command {
...
+                    Arg::new("policies")
+                        .num_args(1..)
+                        .required(true)
+                        .help("list of policies to save")
+                )
+                .arg(
+                    Arg::new("output")
+                    .long("output")
+                    .short('o')
+                    .required(true)
```
flavio, line 543: Being this flag required, it means it's not possible to output the contents of the policy to the STDOUT, like `docker save` does

I'm fine not doing that

src/cli.rs
```diff
@@ -527,6 +527,35 @@ pub fn build_cli() -> Command {
...
+                    .value_name("PATH")
+                    .help("path where the file will be stored")
```
flavio, line 545: This isn't really a path, the user has to provide the name of the file to write to

src/cli.rs
```diff
@@ -527,6 +527,35 @@ pub fn build_cli() -> Command {
...
+                )
+
+        )
+        .subcommand(
+            Command::new("load")
+                .about("load policies from tar file")
+                .arg(
+                    Arg::new("input")
+                        .long("input")
+                        .required(true)
```
flavio, line 555: Also in this case, by being required, it's not possible to read the contents of the policy straight from STDIN

src/load.rs
```diff
@@ -0,0 +1,17 @@
+use anyhow::Result;
+use flate2::read::GzDecoder;
+use policy_evaluator::policy_fetcher::store::Store;
+use std::fs::File;
+use tar::Archive;
+
+// load policies inside the tarball provided by source_path into the default store
+pub(crate) fn load(source_path: String) -> Result<()> {
```
flavio, line 8: Do not use a `String`, but either a `&str` or `&std::path::Path`

src/load.rs
```diff
@@ -0,0 +1,17 @@
...
+    let default_store = Store::default();
+    let destination_path = default_store.root;
+    let tar_gz = File::open(source_path)?;
```
flavio, line 11: I prefer to `map_err` and wrap the error into a `anyhow!("cannot open file {}: {}", source_path, e)`. This gives more clues to the user when something goes wrong.
The errors by itself is pretty dry

src/load.rs
```diff
@@ -0,0 +1,17 @@
...
+    let tar = GzDecoder::new(tar_gz);
+    let mut archive = Archive::new(tar);
+    archive.unpack(destination_path)?;
```
flavio, line 14: same here, I would add some extra context

src/save.rs
```diff
@@ -0,0 +1,23 @@
+use anyhow::Result;
+use flate2::write::GzEncoder;
+use flate2::Compression;
+use policy_evaluator::policy_fetcher::store::{PolicyPath, Store};
+use std::fs::File;
+
+// saves all policies in a tarball with the name provided as output.
+// policies must be inside the default store.
+pub(crate) fn save(images: Vec<&String>, output: &str) -> Result<()> {
+    let tar_gz = File::create(output)?;
```
flavio, line 10: same here, add some context to the error

src/save.rs
```diff
@@ -0,0 +1,23 @@
...
+    let enc = GzEncoder::new(tar_gz, Compression::default());
+    let mut tar = tar::Builder::new(enc);
+
+    for image in images {
+        let store = Store::default();
+        let wasm_path = crate::utils::wasm_path(image.as_str())?;
+        let mut file = File::open(wasm_path)?;
```
flavio, line 17: same here, add more context. This is even more important here because multiple policies are being read, the user must know which one failed

flavio (commented):
Overall ok, aside from my comments I think there are two major things to be discussed:

* docker `save` produces a simple tarball, not a `tar.gz`. This is done on purpose so that the end user can use his favorite compression tool. Because of that, I would just drop the gzip feature you've added
* docker `load` expects a simple tarball as input
* docker `load` and `save` commands worked with `STDIN` and `STDOUT` by default. This allows some scripting using unix pipes. This is used to compress/decompress the tarball with 3rd party compression utilities. I think it would be nice to replicate this behavior

e2e-tests/e2e.bats
```diff
@@ -102,3 +102,23 @@ kwctl() {
     [ "$status" -ne 0 ]
     [[ "$output" == "Error: Wrong value inside of policy's metatada for 'executionMode'. This policy has been created using Rego" ]]
 }
+
+@test "save and load" {
+    kwctl pull registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9
+    kwctl pull https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm
```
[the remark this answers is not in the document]
raulcabello: done https://github.com/kubewarden/kwctl/pull/348/commits/86a1265fef7b862fec9399ae6dda43878d36d9c9

e2e-tests/e2e.bats
```diff
@@ -102,3 +102,23 @@ kwctl() {
...
+    kwctl rm registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9
+    kwctl rm https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm
+    kwctl policies
+    [ "$output" = "" ]
+    kwctl load --input policies.tar.gz
+    kwctl policies
+    [[ $(echo "$output" | wc -l) -eq 6 ]]
+    [[ "$output" =~ "registry://ghcr.io/kubewarden/policies/pod-privileged:v0.1.9" ]]
+    [[ "$output" =~ "https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policy.wasm" ]]
+    rm policies.tar.gz
```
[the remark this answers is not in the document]
raulcabello: https://github.com/kubewarden/kwctl/pull/348/commits/9bc6992d774b5a121123914a3a75b1ad489ca0fd

src/cli.rs
raulcabello: https://github.com/kubewarden/kwctl/pull/348/commits/b5037621fe1e1730ce73087090b321ecfa560486

src/load.rs
raulcabello: https://github.com/kubewarden/kwctl/pull/348/commits/15da194fdf3e2f42b1781e1ecd9d822dd2fb95e5

src/load.rs
raulcabello: make sense. done in https://github.com/kubewarden/kwctl/pull/348/commits/0798f9170407058434b66f02e205dfdf1f92ef7d

src/load.rs
raulcabello: https://github.com/kubewarden/kwctl/pull/348/commits/0798f9170407058434b66f02e205dfdf1f92ef7d

src/save.rs
raulcabello: https://github.com/kubewarden/kwctl/pull/348/commits/0798f9170407058434b66f02e205dfdf1f92ef7d

src/save.rs
raulcabello: https://github.com/kubewarden/kwctl/pull/348/commits/0798f9170407058434b66f02e205dfdf1f92ef7d

src/cli.rs
raulcabello: We will just accept tar.gz files as agreed. Added to help message here https://github.com/kubewarden/kwctl/pull/348/commits/b1ca9ffbc94ee85ba0565cd4ca475ac4352cc69e

e2e-tests/e2e.bats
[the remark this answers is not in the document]
flavio: nice job, thanks!

src/cli.rs
```diff
@@ -527,6 +527,35 @@ pub fn build_cli() -> Command {
                         .help("Path to a Docker config.json-like path. Can be used to indicate registry authentication details")
                 )
         )
+        .subcommand(
+            Command::new("save")
+                .about("save policies in tar.gz file")
```
flavio, line 532:
```suggestion
                .about("save policies to a tar.gz file")
```

flavio (approved):
LGTM, I left a minor fix

viccuad (approved):
LGTM!

PR #354, opened 2022-11-15
Save and load scripts

raulcabello:
These scripts are required for airgap instalation. 

- `kubewarden-save-policies.sh` saves all policies in the file provided inside a tar.gz
- `kubewarden-load-policies.sh` load policies inside a tar.gz 

Steps to test it:

policies.txt example
```
registry://ghcr.io/kubewarden/policies/capabilities-psp:v0.1.9
registry://ghcr.io/kubewarden/policies/psp-capabilities:v0.1.5
https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.6/policy.wasm

```

- Download kwctl 1.3.2
- Create a private registry
-  .`/kubewarden-save-policies.sh --policies-list policies.txt --policies policies.tar.gz`
- remove all policies in `policies.txt` with `kwctl rm`
- `./kubewarden-load-policies.sh --policies policies.tar.gz --policies-list policies.txt --registry localhost:5000 --sources-path sources.yml`
- verify that the following policies are present in the private registry:
```
localhost:5000/kubewarden/policies/capabilities-psp
localhost:5000/kubewarden/policies/psp-capabilities
localhost:5000/kubewarden/pod-privileged-policy/releases/download/v0.1.6/policy.wasm
```

Fix https://github.com/kubewarden/kubewarden-controller/issues/344


kubewarden-load-policies.sh
```diff
@@ -0,0 +1,85 @@
...
+policies=()
+while read policy; do
+    policies+=("${policy}");
+done < ${list}
+
+for policy in "${policies[@]}"; do
+    if [[ $policy == registry://* ]]; then
+       oldPolicyUrl=$(awk -Fregistry:// '{print $2}' <<< $policy)
+       oldRegistry=$(echo "$oldPolicyUrl" | cut -f1 -d"/")
+       newPolicyUrl="registry://${oldPolicyUrl/$oldRegistry/$registry}"
```
flavio, line 76: I'm getting lost here. Maybe some inline docs could help?

kubewarden-load-policies.sh
```diff
@@ -0,0 +1,85 @@
...
+       pushPolicy $newPolicyUrl
+    fi
+    if [[ $policy == https://* ]]; then
+       oldPolicyUrl=$(awk -Fhttps:// '{print $2}' <<< $policy)
+       oldPolicyUrlWithoutDomain=$(awk -F/ '{print $2}' <<< $policy)
+       newPolicyUrl="registry://$registry/${oldPolicyUrl#*/}"
```
flavio, line 82: Same here

flavio (commented):
Overall OK. I would however like to have a GH action that does the following steps:

* start a docker registry
* use the save script to download some test policies into a tar.gz
* use the load script to push the saved policies into the local test registry
* ensure (maybe via `crane`) that each policy has been pushed to the local test registry

kubewarden-load-policies.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
```
viccuad, line 1: I would add `set -euo pipefail`. Given that we are introducing shell scripts, I would add a shellcheck CI job as we have in other repos.

kubewarden-load-policies.sh
```diff
@@ -0,0 +1,85 @@
...
+        kwctl push $policy $newPolicyUrl
+    fi
+}
+
+while [[ $# -gt 0 ]]; do
+    key="$1"
+    case $key in
+        -r|--registry)
+        registry="$2"
+        shift # past argument
```
kravciak, line 28: You can move one `shift` after `key="$1"`, out of case block to not repeat it for every argument.
Or use `shift 2` if you are shifting twice.

kubewarden-load-policies.sh
```diff
@@ -0,0 +1,85 @@
...
+        usage
+        exit 1
+        ;;
+    esac
+done
+if [[ -z $registry ]]; then
+    usage
+    exit 1
+fi
+if [[ $help ]]; then
```
kravciak, line 60: This produces error if variable does not exist (without -h param). If you want to see if it's set use `-v`
```suggestion
if [[ -v help ]]; then
```

scripts/kubewarden-load-policies.sh
```diff
@@ -0,0 +1,86 @@
...
+        usage
+        exit 1
+        ;;
+    esac
+done
+if [[ -v help ]]; then
+    usage
+    exit 0
+fi
+if [[ -z $registry ]]; then
```
kravciak, line 62: Prevents ugly error if variable is not set
```suggestion
if [[ -z ${registry:-} ]]; then
```

kubewarden-load-policies.sh
raulcabello: not sure if that doesn't work, or if I'm missing something (I'm not an expert in bash :sweat_smile: ) I took this code from the `rancher-save-images.sh` script which uses the same switch.

I get the following error:
`./scripts/kubewarden-load-policies.sh: line 43: $2: unbound variable`
I think this is because shift needs to happens after `policies="$2"` and not before as it happens if I move the shift out of the switch block after key="$1".

kubewarden-load-policies.sh
raulcabello: done in https://github.com/kubewarden/kwctl/pull/354/commits/1eb07dd70a233cdc5203b2befa13a2e5e8c10b89

kubewarden-load-policies.sh
raulcabello: done in https://github.com/kubewarden/kwctl/pull/354/commits/1eb07dd70a233cdc5203b2befa13a2e5e8c10b89 and https://github.com/kubewarden/kwctl/pull/354/commits/25e00063b214877600c5ff9464b7c89d34bd3816

kubewarden-load-policies.sh
raulcabello: https://github.com/kubewarden/kwctl/pull/354/commits/94f35ea12fdf8501600729f56121283b58c99c56

kubewarden-load-policies.sh
raulcabello: https://github.com/kubewarden/kwctl/pull/354/commits/94f35ea12fdf8501600729f56121283b58c99c56

kubewarden-load-policies.sh
raulcabello: done [here](https://github.com/kubewarden/kwctl/pull/354/commits/20b0c57ea52b1df5d46cd7cf661ccae8d4969c77). I needed to change `policies="$2"` to `policies="$1"` as you suggested. Thanks

scripts/kubewarden-save-policies.sh
```diff
@@ -0,0 +1,55 @@
...
+        policies="$1"
+        shift # past value
+        ;;
+        -l|--policies-list)
+        list="$1"
+        shift # past value
+        ;;
+        -h|--help)
+        help="true"
+        shift
```
kravciak, line 28: There are no params to shift, key is already shifted
```suggestion
```

kravciak (approved):

.github/workflows/airgap-test.yml
```diff
@@ -0,0 +1,44 @@
+name: Airgap
+on:
+  - push
+  - pull_request
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  airgap-e2e-test:
```
viccuad, line 10: I would prefer to implement this as part of the existing `make e2e-test` target or a new `make e2e-airgap`, so it can be run locally.
We can leave "install kwctl" as a GHA step.

flavio: I think it's better to leave that here. Running the test is cheap and we end up doing that on each PR

.github/workflows/airgap-test.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: |
+          export CONTAINER_ID=$(docker run -d -p 5000:5000 --name registry registry:2) 
+          echo "CONTAINER_ID=${CONTAINER_ID}" >> $GITHUB_ENV
+      - name: Install kwctl
+        run: cargo install --path .
+      - name: Save policies
+        run: ./scripts/kubewarden-save-policies.sh --policies-list e2e-tests/airgap/policies.txt --policies policies.tar.gz
+      - name: Remove policies from store
+        run: |
+          kwctl rm registry://ghcr.io/kubewarden/policies/capabilities-psp:v0.1.9
```
flavio, line 25: Please, don't use an official policy. This would influence our usage stats

flavio (changes requested):
Overall ok, I just want the test to not influence our download stats

.github/workflows/airgap-test.yml
raulcabello: I would leave it as it is as there are a few things that work in a different way than the current `e2e-test`:
- We need to run a local registry. GHA is creating the registry with a docker image. 
- Scripts are using [kwctl installed](https://github.com/raulcabello/kwctl/blob/save-load-script/scripts/kubewarden-save-policies.sh#L44) in your machine unlike e2e test that uses cargo run. 

I don't expect many changes to airgap in the future once it is finished, and GHA could be run locally with https://github.com/nektos/act

.github/workflows/airgap-test.yml
raulcabello: yes, you are totally right! done in https://github.com/kubewarden/kwctl/pull/354/commits/43440e53145a57611ac80f1651c50d9bb982704a

.github/workflows/airgap-test.yml
viccuad: let's leave it as it is then :).

viccuad (approved):

Issue #54, opened 2021-06-23
Implement benchmark subcommand


flavio:
It would be nice to allow policy authors to benchmark their policies. This would be useful to spot how the source language influences the final evaluation time of the policies.

Using regular benchmark solutions won't work, because they would start kwctl multiple times. On each invocation `kwctl` would have to load and optimize the Wasm module. This is an initial performance hit that goes away once the policy is "hot" in memory (as it happens with policy-server).

## Acceptance criteria

* kwctl has a `bench` sub-command
* The uses the same UX of the `run` command, but it has some extra flags about number of iterations and other things specific to the benchmark operation.
* The extra flags (compared to `run`) should have default values. This is to ensure `kwct run` and `kwctl bench` can be swapped out inside of scripts

flavio:
@srenatus provides some rough code we might want to reuse :bow: 

```rust
use criterion::{black_box, criterion_group, criterion_main, Criterion};
use policy_evaluator::policy_evaluator::PolicyEvaluator;
use anyhow::Result;
use std::process;
use std::fs::File;
use std::path::Path;
use std::io::BufReader;
use tracing::Level;
use tracing_subscriber::FmtSubscriber;
struct Req {
    file: String,
    allow: bool,
}
fn bench_group(c: &mut Criterion, dir: String, reqs: Vec<Req>) {
    let mut group = c.benchmark_group(dir.to_string());
    for m in ["policy.wasm", "kw_policy.wasm"].iter() {
        for r in reqs.iter() {
            group.bench_function(format!("{}/{}/{}", *m, r.file, r.allow), |b| {
                let request = read_request_file(&format!("{}/{}", dir.to_string(), &r.file)).unwrap();
                let mut policy_evaluator = black_box(PolicyEvaluator::new(&Path::new(&format!("{}/{}",  dir.to_string(), m)), None).unwrap());
                // run once before benchmarking
                let vr = policy_evaluator.validate(request.clone());
                if r.allow != vr.allowed {
                    fatal_error(format!("unexpected result from {}: {:?}", m, vr))
                }
                b.iter(|| policy_evaluator.validate(request.clone()));
                // run once more, see if we're still reporting the correct result
                let vr_after = policy_evaluator.validate(request.clone());
                if r.allow != vr.allowed {
                    fatal_error(format!("unexpected result from {} (after benchmark): {:?}", m, vr_after))
                }
            });
        }
    }
    group.finish();
}
[...]

PR #368, opened 2022-11-25
add bench command

flavio:
Add the "bench" command. This can be used to benchmark how the `validate_settings` and the `validate` functions of a policy perform.
    
This can be useful for end developers to understand the performance of their WebAssembly modules.

Fixes https://github.com/kubewarden/kwctl/issues/54

From the CLI flags POV, the `bench` command is a superset of the `run` command. Because of that the `run` code has been refactored a bit to allow code reuse.

While working on the e2e tests I realized we were pulling all our policies from the "official" sources instead of the "test" one. This was screwing our download stats and I fixed that.

# Example

This is how a local policy can be run with the usual `run` command:

```console
kwctl run \
  --sources-path sources.yaml \
  -s ../verify-image-signatures/settings-pod_signed_with_cert.yaml \
  -r ../verify-image-signatures/pod_signed_with_cert.json \
  ../verify-image-signatures/policy.wasm
```

This produces the following output:

```console
{"uid":"1299d386-525b-4032-98ae-1949f69f9cfc","allowed":true,"patchType":"JSONPatch","patch":"W3sib3AiOiJyZXBsYWNlIiwicGF0aCI6Ii9zcGVjL2NvbnRhaW5lcnMvMC9pbWFnZSIsInZhbHVlIjoicmVnaXN0cnktdGVzdGluZy5zdmMubGFuL2t1YmV3YXJkZW4vcG9kLXByaXZpbGVnZWQ6djAuMS45QHNoYTI1NjpmMTE0M2VjMjc4NmUxM2Q3ZDMzMzVkYmI0OTg1Mjg0MzhkOTEwNjQ4NDY5ZDNmMzk2NDdlMWNkZTY5MTRkYThkIn0seyJvcCI6ImFkZCIsInBhdGgiOiIvYXBpVmVyc2lvbiIsInZhbHVlIjoidjEifSx7Im9wIjoiYWRkIiwicGF0aCI6Ii9raW5kIiwidmFsdWUiOiJQb2QifV0=","auditAnnotations":null,"warnings":null}
```

The bench command can be invoked replacing `run` with `bench`:

```console
kwctl bench \
  --sources-path sources.yaml \
  -s ../verify-image-signatures/settings-pod_signed_with_cert.yaml \
  -r ../verify-image-signatures/pod_signed_with_cert.json \

```

This produces the following output:

```console
validate_settings warming up for 3.00s
validate_settings mean warm up execution time 20.05µs running 252.5 thousand iterations
validate_settings [252.5 thousand iterations in 5.07s with 100.0 samples]:
	elapsed	[min mean max]:	[19.84µs 20.20µs 28.65µs] (sample data: med = 20.03µs, var = 852.20µs², stddev = 923.14ns)
validate warming up for 3.00s
validate mean warm up execution time 241.32µs running 25.2 thousand iterations
validate [25.2 thousand iterations in 5.89s with 100.0 samples]:
	elapsed	[min mean max]:	[215.60µs 233.11µs 276.45µs] (sample data: med = 235.53µs, var = 133.94ms², stddev = 11.57µs)
```

The output is colored in a fancy way too :)


viccuad (approved):
LGTM! nice addition

src/cli.rs
```diff
@@ -527,6 +527,157 @@ pub fn build_cli() -> Command {
...
+                    .number_of_values(1)
+                    .value_name("NUM")
+                    .help("How many resamples should be done")
+                )
+                .arg(
+                    Arg::new("num_samples")
+                    .long("num-samples")
+                    .number_of_values(1)
+                    .value_name("NUM")
+                    .help("Recommended at least 50, above 100 doesn’t seem to yield a significantly different result")
```
jvanz, line 552: Even if the name of the argument helps to understand what it does, I think we should add a brief explanation what it means. Something similar of what we do in the `warm_up_time` argument.

jvanz (approved):

src/cli.rs
flavio: done

PR #375, opened 2022-12-02
Improve airgap workflow

kravciak:
## Description

Airgap workflow is currently cumbersome, this is an attempt to improve it

This combines 4 scripts for saving & loading images and policies into one.
Also adding defaults for parameters that can be guessed from other files.

## New Workflow:
```bash
# Generate list of dependencies required by airgap. Save in file for other commands:
airgap list | tee file.json

# Pull images and policies and save them to archives in current directory:
airgap pull --list file.json

# Push images and policies from created archives to local registry. Run in DRY mode:
airgap push --list file.json --registry=localhost:5123 --dry --insecure

# Helm install charts from local files. Registry is used for recommendedPolicies setup:
airgap install --list file.json --registry 172.18.0.4:5000 --insecure

```


viccuad:
Just as a heads up, given that it will change the workflow, this PR should be together with one that updates the docs (analogous to https://github.com/kubewarden/docs/pull/150) and one that updates the blogpost (analogous to https://github.com/kubewarden/kubewarden.io/pull/153).

jvanz (changes requested):
This is a very nice script! One script for everything is very nice idea. 

I like that the script download the list if the user does not define the path to the file. But if I define some of the lists (images or policies) I would prefer that the script does not download the other file automatically. I think that if I define the list I do not want the script to download another file without my command.

scripts/airgap.sh
```diff
@@ -0,0 +1,195 @@
+#!/usr/bin/env bash
+set -eEuo pipefail
+
+script=$(basename $0)
+usage() {
+    echo "Usage:
+      $script action --parameters
+      $script pull            [--policy-list policies.txt] [--image-list images.txt] [--policies policies.tar.gz] [--images images.tar]
+      $script push --registry [--policy-list policies.txt] [--image-list images.txt] [--policies policies.tar.gz] [--images images.tar] [--kwctl-sources sources.yaml]
```
viccuad, line 9:
```suggestion
      $script push --registry myregistry:port [--policy-list policies.txt] [--image-list images.txt] [--policies policies.tar.gz] [--images images.tar] [--kwctl-sources sources.yaml]
```

scripts/airgap.sh
```diff
@@ -0,0 +1,195 @@
...
+# Functions
+
+# Print green colored message
+function step() { printf -- "\e[32m${*}\e[0m\n"; }
+
+# Get latest policiy and image lists from artifacthub
+# Use text files from kubewarden-defaults chart
+function pull_latest_lists() {
+   tag=$(helm search hub kubewarden-defaults -o json | jq -er '.[] | select(.repository.url=="https://charts.kubewarden.io").version')
+   assets_urls=$(curl -sL https://api.github.com/repos/kubewarden/helm-charts/releases/tags/kubewarden-defaults-$tag | jq -er '.assets[].browser_download_url')
```
viccuad, line 111: I don't like this, the canonical locations of artifacts are either the HTPPS charts.kubewarden.io with unsigned charts, or the OCI registry ghcr.io/kubewarden/* with signed helm charts. GH releases are indeed being consumed by https://charts.kubewarden.io/index.yaml, but I'm not sure we should consume them directly.

Also, given that air gap users are more security-conscious than the average, I expect them to verify the signatures of the helm charts prior to obtaining the images and policies list. Hence why in the current docs there's a "stop gap" between obtaining the helm charts (as the user desires, and verifying them or not), and then using an image list (either from the helm charts or the release).

I would prefer to do a `helm pull` either from one repo or the other, specifed by the user. Then, have a script command that finds the lists and compiles them.

Also note, that the current approach doesn't use the GH release files ala
`kubewarden-crds-1.2.3_kubewarden-defaults-1.2.8_kubewarden-controller-1.2.8_policylist.txt`. I think not using them is indeed more elegant, wonder what the rest think.

scripts/airgap.sh
```diff
@@ -0,0 +1,195 @@
...
+   # if source is url download it
+   if [[ "$source" =~ ^https:// ]]; then
+      curl -sL $source -o "$tmpfile"
+      source=$tmpfile
+   fi
+
+   grep -qF '://' $source && command=kwctl || command=docker
+   cat $source | while read line || [[ -n $line ]]; do
+      echo "Pull: $line"
+      $command pull "$line" > /dev/null
```
viccuad, line 134:
```suggestion
      $command pull "$line"
```

I am happy seeing the list of things being pulled; ideally the script is run only once, can be verbose.

scripts/airgap.sh
```diff
@@ -0,0 +1,195 @@
...
+   cat $source | while read image || [[ -n $image ]]; do
+      echo -n "Push: "
+      if [ $command == 'docker' ]; then
+         local target=$REGISTRY/${image#*/}
+         docker tag $image $target
+         docker push -q $target
+      elif [ $command == 'kwctl' ]; then
+         local target=$(sed -E "s#.*://[^/]+#registry://$REGISTRY#" <<< "$image")
+         echo "$target"
+         kwctl push $image $target ${kwctl_args:-} >/dev/null
```
viccuad, line 167:
```suggestion
         kwctl push $image $target ${kwctl_args:-}
```

I would like to print the list of policies being pushed to be fair.

viccuad (commented):
It needs some work on shellcheck lint errors, see https://github.com/kubewarden/kwctl/actions/runs/3602159868/jobs/6068780134.

I like where it's going, but I would prefer for it to be even simpler somehow. Maybe less options. My concern is that if I were the one setting the air gap, I would end up reviewing the script for security reasons. And the others were more linear in execution and to the point, so the review was easier.

it would also be nice to update the airgap [tests run](https://github.com/kubewarden/kwctl/actions/runs/3602159868/jobs/6068780350) (see https://github.com/kubewarden/kwctl/blob/main/.github/workflows/airgap-test.yml) to use the new script. I suppose the PR run will not find the correct script to checkout, but that's ok, at least it informs of the usage of the script.

scripts/airgap.sh
```diff
@@ -0,0 +1,274 @@
...
+Parameters:
+  -l|--list       Json file with list of required dependencies
+  -r|--registry   Private registry where artifacts are pushed. Helm charts are configured for this registry
+  --dry           Do not perform actions, only print commands that would be executed
+  --cert-manager  Include cert-manager dependencies in the generated list
+  --insecure      For pull command skip kwctl certificate verification (an alternative to sources.yaml file)
+                  For install disable policyServer verification of registry
+
+Commands:
+  list     Generate json list of dependencies. You can redirect it to a file
```
viccuad, line 43: I'm considering if it isn't better to drop the list command; it is difficult to consider all the options that would go here:  which versions? which values.yaml, that can influence the images being deployed? which charts (currently missing kubewarden-controller as I see it)? from which repos do we pull the charts? are they already mirroring the kubewarden charts in their own repo? do they prefer to do `helm pull` + check signatures beforehand instead?

I would just tell the users to obtain (and verify) the kubewarden helm charts on their own in the docs (they would do `helm pull`).

Once they have the kubewarden charts via `helm pull` or similar, untar them to get the `imagelist.txt` and `policylist.txt`, do `./airgap.sh pull`, `./airgal.sh push`, `./airgap.sh scaffold-install` (see below) to create values.yaml, and then commit the files to git or do `helm install` as normal docs.

This exact same process can be used for Kubewarden upgrades by users, just substitute the last step for `helm upgrade -i`.

scripts/airgap.sh
```diff
@@ -0,0 +1,274 @@
...
+# Defaults
+IMAGES_TAR="kubewarden-images.tar"
+POLICIES_TAR="kubewarden-policies.tar.gz"
+
+PROG=$(basename "$0")
+USAGE="\
+Usage:  $PROG command [--parameters]
+
+  $PROG list [--cert-manager]
+  $PROG pull -l file.json [--dry]
```
viccuad, line 30: I wonder about using json as the image list format, do we know if it would be useful for any specific tool? 
Maybe is easier to just have a line list as it was. It also makes it as the output used in similar tools by the community, such as the [helm image plugin](https://github.com/cvila84/helm-image).

scripts/airgap.sh
```diff
@@ -0,0 +1,274 @@
...
+  --dry           Do not perform actions, only print commands that would be executed
+  --cert-manager  Include cert-manager dependencies in the generated list
+  --insecure      For pull command skip kwctl certificate verification (an alternative to sources.yaml file)
+                  For install disable policyServer verification of registry
+
+Commands:
+  list     Generate json list of dependencies. You can redirect it to a file
+  pull     Pull components from list to current directory
+  push     Push components in list from current directory to local registry
+  install  Install helm charts configured with local registry
```
viccuad, line 46: While I like the idea of providing a command for installing the helm charts, I think it adds a new layer of indirection for users, as they aren't aware of what exactly has happened there, and they need to read outputs.
Also, I would expect people to use GitOps to deploy their Helm charts to the cluster. They would just commit `values.yaml`, `insecure-sources.yaml` and similar into git, and then call an install job.

Maybe instead of having an `install` subcommand that calls `helm install`, we can provide a `scaffold-install` subcommand, that creates a values.yaml and insecure-sources.yaml or so as needed. And then the user can review them, change them if needed, and make the call to `helm install` or `upgrade -i` themselves with whatever tooling they have. This gives them the correct values.yaml pointing the internal registry, disabling the recommendedPolicies, and so, and they can review it and be aware of what are they are going to get. We could even document the scaffolded values.yaml with the reasons; for example "disabling recommendedPolicies, you can re-enable them if you have mirrored the wasm modules in your air-gapped registry".

scripts/airgap.sh
```diff
@@ -0,0 +1,274 @@
...
+  - kubewarden helm repository
+
+Examples:
+  Generate list of dependencies required by airgap. Save in file for other commands:
+    $PROG list --cert-manager | tee file.json
+
+  Pull images and policies and save them to archives in current directory:
+    $PROG pull --list file.json
+
+  Push images and policies from created archives to local registry. Run in DRY mode:
```
viccuad, line 59:
```suggestion
  Push images and policies from created archives to local registry. Run in dry-run mode:
```

(DRY on is own, and in uppercase could be misunderstood as [Don't Repeat Yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself))

scripts/airgap.sh
```diff
@@ -0,0 +1,274 @@
...
+function command_list {
+   local tag url urls images policies charts
+
+   tag=$(helm search repo kubewarden/kubewarden-defaults -o json | jq -er 'first.version')
+   urls=$(curl "https://api.github.com/repos/kubewarden/helm-charts/releases/tags/kubewarden-defaults-$tag" | jq -er '.assets[].browser_download_url')
+
+   # images
+   url=$(grep 'images.txt$' <<< "$urls" | tail -1)
+   images=$(curl "$url" | to_json 'images')
+   if [ -v CERTMAN ]; then
```
viccuad, line 103: I see where you are going by taking care of cert-manager in list and install subcommands, but I think it's better to leave cert-manager out of this script and have it directly in our CI tests, it would be super helpful there.

The current code is taking the images the latest cert-manager helm chart, and installing the latest helm chart, but that may or may not be desirable. Also the users may need to configure cert-manager in their own way for example. And by configuring a helm chart, you may change the usage of images, too. I think its ok to opt-out to all of this, and refer users to upstream cert-manager for its airgapped deployment.

viccuad (commented):
Did a review, it's looking good, and I think you are touching into even more things to take into account, so added some comments to reduce scope.

About the "helm charts vs release canonical location of artifacts", let me see if I can explain it better:

Right now the charts are being published duplicated. We have:
- A. https://charts.kubewarden.io/index.yaml, a typical https helm repo, with an index.yaml file that lists where to download the tars from. These are unsigned. The tars are hosted in GH releases, but this is an implementation detail, and it can change at any moment. We could put them in AWS and update the index.yaml to point to those AWS urls, or if someone builds a product out of Kubewarden. Hence we shouldn't depend on them. 
- B. As a OCI artifacts, under the OCI ghcr registry, `ghcr.io/kubewarden/charts/`. Example: https://github.com/kubewarden/helm-charts/pkgs/container/charts%2Fkubewarden-crds.
  This ones are signed via Sigstore, and OCI registries are supported by helm since helm 3.8. Since OCI registries don't have an index (yet), they aren't searchable via `helm repo search` though.

Also, users of airgap systems may mirror the helm charts themselves, in their own Helm repositories (which may very well be OCI ones, which is what everyone is doing nowadays). kubewarden may be shipped with products too, such as Rancher, where that happens (under rancher/helm-charts, or [Rancher Federal](https://rancherfederal.github.io/carbide-docs/docs/registry-docs/za-enforcement)).

Since we need to support all those cases, and allow for users to use case B and verify the sigstore signatures of the charts, I think it's better to leave the obtaining of the charts out of the scope of the script.

flavio:
Closing, we will revisit that later as discussed with @kravciak

Issue #360, opened 2022-11-22
kwctl does not honor trusted certificate authorities on the system


Martin-Weiss:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

kwctl pull against on-premise registry with self-signed certificate / certificate authority fails with:

`kwctl pull registry://registry01.suse:5000/rke-prod/ghcr.io/kubewarden/policies/user-group-psp:v0.2.0
`
```
2022-11-22T15:54:48.099449Z  WARN rustls::conn: Sending fatal alert BadCertificate    
Error: the policy registry://registry01.suse:5000/rke-prod/ghcr.io/kubewarden/policies/user-group-psp:v0.2.0 could not be downloaded due to error: error sending request for url (https://registry01.suse:5000/v2/): error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
```

### Expected Behavior

The tool should trust the CA of the systems ca-bundle.pem.

A cli parameter to specify the trusted CA bundle should be available.

### Steps To Reproduce

Use on premise registry for policies and try to use pull / push with kwctl.

### Environment

```markdown
- OS: SLES 15 SP4
- Architecture: x86_64
```

Martin-Weiss:
Yes - there seems to be a way to use the sources.yaml with kwctl but here I can not specify the certificate of the CA in a way that works:

```
source_authorities:
 "registry01.suse:5000":
 - type: Data
   data: |
    -----BEGIN CERTIFICATE-----
    MIIDhzCCAm+gAwIBAgIUVS8hv/sK6r15ZjVf/Zz/AQdoqjAwDQYJKoZIhvcNAQEL
    BQAwUzELMAkGA1UEBhMCREUxGzAZBgNVBAgMEkJhZGVuLVd1ZXJ0dGVtYmVyZzEN
    MAsGA1UECgwEU1VTRTEYMBYGA1UEAwwPcmVnaXN0cnkwMS5zdXNlMB4XDTIxMDMy
    NDE0MTMzM1oXDTMxMDEzMTE0MTMzM1owUzELMAkGA1UEBhMCREUxGzAZBgNVBAgM
    EkJhZGVuLVd1ZXJ0dGVtYmVyZzENMAsGA1UECgwEU1VTRTEYMBYGA1UEAwwPcmVn
[...]

PR #386, opened 2022-12-14
fix: use system certificates

flavio:
Make kwctl use the system certificates, instead of relying on the ones bundled by rustls.

Fixes https://github.com/kubewarden/kwctl/issues/360


viccuad (approved):
LGTM! tested locally already, for `registry://` and `https://`

PR #408, opened 2023-01-19
Run continuous integration on every change.

jvanz:
## Description

Recently we just notice an issue in the kwctl build due a lib update on release. To spot this kind of problem earlier this commit changes when we run the continuous integration jobs. With this change the code will be verified and build on every change.


flavio:
This will build the kwctl binaries every time something lands into the `main` branch. Do I get it right?

jvanz:
> This will build the kwctl binaries every time something lands into the `main` branch. Do I get it right?

Yes, that's right.

jvanz:
I've updated the CI jobs to run basic source code checks (clippy, rustfmt and check) in the pull requests. Just to get some basic validation in the PR as we do so far.

jvanz:
I've also added another job to check if the `Cargo.toml` version match with the version the pushed tag.

.github/workflows/cargo-file-checks.yml
```diff
@@ -0,0 +1,24 @@
+name: Cargo file checks
```
jvanz, line 1: If you think this is a good idea. After merging this PR I can create a github action to use the same code in other rust repositories.

flavio (approved):
cool, thanks!

jvanz:
@flavio the `e2e tests` CI check will not get green because I've change the file name and where the jobs are run. Therefore, you can merge the PR ( I cannot do it. I do not have the permissions for that )

PR #422, opened 2023-02-07
fix: scaffold should not generate the backgroundAudit field

flavio:
Currently the `backgroundAudit` field is not known by the Kubewarden's stable CRDs.

Because of that, we should not generate this attribute at scaffold time.

This is a quick workaround for https://github.com/kubewarden/kubewarden-controller/issues/395

Note: the field is not scaffolded when it's `true`, which is the default case. A policy with `backgroundAudit` set to `false` would still cause the scaffold code to produce this field, leading to this bug again. However this is not a big deal, since no policy is currently setting this attribute. On the other hand, we need kwctl to generate this attribute for our development purposes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4108,9 +4108,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.91"
+version = "1.0.92"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


flavio:
Once merged, I'll tag a patch release of kwctl

jvanz (approved):

Issue #429, opened 2023-02-16
kwctl `scaffold`: handle the `contextAwareResources` attribute


flavio:
As a result of the [context aware RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0015-context-aware-policies.md), policies will now feature a new metadata attribute called `contextAwareResources`. Moreover, the `ClusterAdmissionPolicy` Custom Resource will feature a new `contextAware` attribute.

The `scaffold` command has to be updated so that it generates the new `contextAware` attribute of `ClusterAdmissionPolicy` resources. The data used by `scaffold` will be taken from the policy metadata.

However, it's important to ensure the Kubernetes administrator is comfortable giving access to these resources.

Because of that, when scaffolding a `ClusterAdmissionPolicy` resource starting from a context aware policy, we will by default exit with an error. We can show something like that:

```console
ERROR: the policy requires access to Kubernetes resources at evaluation time. This could pose a security risk, hence you have
to carefully review the list of resources the policy intends to access.

This is the list of resources it will have access to:

< data from the `contextAwareResources` metadata> 

If granting access to these resources is legit, re-run this command adding the `--allow-context-aware` flag
```

When the `allow-context-aware` boolean flag is provided, the `scaffold` command will use the data inside of `contextAwareResources` to populate the `ClusterAdmissionController CRD.
We should print a warning message on the `stderr` (not the `stdout`, this would break a use case like `kwctl scaffold | kubectl apply -`). The warning message could be something along these lines:

```console
WARNING: this policy has been granted access to Kubernetes resources at evaluation time. Please review carefully the resources being accessed by the policy.
```

flavio:
Done via https://github.com/kubewarden/kwctl/pull/431

Issue #428, opened 2023-02-16
kwctl `run`: handle new context aware policies


flavio:
We have formalized how context aware policies have to be handled with [this new RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0015-context-aware-policies.md).
As a result of that, we have to change how `kwctl run` handles context aware policies.

By default, when running a context aware policy, kwctl will not allow the policy to access Kubernetes resources.
The user can allow access to Kubernetes resources by using a new boolean flag `--allow-context-aware`. When this flag is enabled the policy will be granted access only to the resources mentioned inside of its metadata (the new `contextAwareResources` data).

## Admission criteria

When running a policy that has `contextAwareResources` inside of its metadata:

* By default:
  * [x] Do not allow access to any Kubernetes resource (basically run the policy with an empty allow list)
  * [x] Print a warning message stating that the policy uses context aware data, but it will not have access to it. Hint the usage of the `--allow-context-aware` flag
* When using the `--allow-context-aware` flag:
  * [x] Print a big warning stating which Kubernetes resources the policy will have access to
  * [x] Allow access only to the resources mentioned inside of the `contextAwareResources` field of policy's metadata

flavio:
Done via https://github.com/kubewarden/kwctl/pull/431

Issue #427, opened 2023-02-16
Update policy metadata to handle new `contextAwareResources` attribute


flavio:
As described by the  [context aware policies RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0015-context-aware-policies.md), the policies that want to access Kubernetes resources at evaluation time must indicate them inside of their metadata.

The policy metadata has to be extended to store this new attribute:

```yaml
contextAwareResources:
# - apiVersion: <apiVersion>
#   kind: <kind>
- apiVersion: v1
  kind: Pod
- apiVersion: apps/v1
  kind: Deployment
- apiVersion: networking.k8s.io/v1
  Kind: Ingress
```

Currently, the policy metadata has already a `contextAware` attribute, which has a boolean value. After this change we will ignore the legacy `contextAware` attribute, both when reading the metadata and when annotating the policy.

## Admission criteria

* [x] `annotate` command must write the `contextAwareResources` attribute
* [x] `annotate` command will stop writing the `contextAware` attribute
* [x] `inspect` command will read only the `contextAwareResources` attribute. A policy that has an empty `contextAwareResources` attribute (or doesn't have it) will be considered as NOT context aware

flavio:
Done via https://github.com/kubewarden/kwctl/pull/431

PR #431, opened 2023-02-22
Context aware

flavio:
This PR adds support of the new context aware feature to kwctl.

Each commit is standalone, and addresses a dedicated issue:

- feat: handle context aware resources attribute - https://github.com/kubewarden/kwctl/issues/427
- fix: better handling of boolean flags
- feat: new context aware implementation - run https://github.com/kubewarden/kwctl/issues/428
- chore(deps): update policy-evaluator dependency
- feat: context aware and scaffold command https://github.com/kubewarden/kwctl/issues/429

Currently the code is built against the `context-aware` feature branch of `policy-evaluator`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -465,9 +465,9 @@ dependencies = [
 
 [[package]]
 name = "bstr"
-version = "1.2.0"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b7f0778972c64420fdedc63f09919c8a88bda7b25135357fd25a5d9f3257e832"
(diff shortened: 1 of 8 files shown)


viccuad (commented):
LGTM! I suppose we should add some e2e tests, at least for `kwctl scaffold` without `--allow-context-aware`, and for `kwctl run` with `--allow-context-aware`.

flavio:
@viccuad yes, you're right. I've added some unit tests to the `scaffold` code. They make the e2e no longer necessary IMHO

flavio:
@viccuad now I've added some more tests for the `run` code. I think we can avoid to touch the e2e tests further.

Also, consider that we would need to create a k8s cluster, add some resources to it and then have a dedicated policy that tests that. This is doable, but not something that needs some work, especially to ensure the tests run fine on local machines.

Issue #418, opened 2023-01-25
Add new `kwctl scaffold artifacthub-pkg` command


viccuad:
Relates to 

As specified in https://github.com/kubewarden/rfc/pull/15, add a new `kwctl scaffold artifacthub-pkg` that generates an `artifacthub-pkg.yml` for a given policy.

## Acceptance criteria

- Do this by implementing a serializable `ArtifacthubPkg{}` struct that matches [artifacthub-pkg.yml](https://togithub.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-pkg.yml). It doesn't need to implement all fields, only the ones we cover. Double check with the ones already written by hand to see we aren't missing any derivable info from metadata.yml. As an example: [verify-image-signatures/blob/main/artifacthub-pkg.yml](https://github.com/kubewarden/verify-image-signatures/blob/main/artifacthub-pkg.yml) 
- Provide a `TryFrom()` from `Metadata{}` for `ArtifacthubPkg{}`, that derives and checks the fields according to the RFC.
- Add a new `kwctl scaffold artifacthub-pkg --version v0.0.0 --usage Usage.md --metadata-path metadata.yml`. This command will print the resulting `artifacthub-pkg.yml`, failing according to the TryFrom definitions in the RFC.
- Provide tests

flavio:
The `scaffold artifacthub` command should also have a cli flag that points to the `questions.yml` file.

I also think the `scaffold artifacthub` should not have a `usage` flag, that's because the policy usage instructions are not part of the `artifacthub-pkg.yml`.

To recap, this is what I would propose as flags of the `scaffold artifacthub` command:

* `--questions-ui QUESTIONS`: mandatory, this will read the contents of the question file
* `--version VERSION`: mandatory

About the `--version` flag. I would start with a mandatory version, next we can make it optional and have the code figure out what is the version of a Rust policy by looking at the contents of the `Cargo.toml`. With other types of policies (like Go, Rego,...) the lack of this flag would lead to an error

PR #436, opened 2023-03-01
feat: Add `kwctl scaffold artifacthub` command, add `kwctl annotate --usage` flag

viccuad:
## Description

Fix https://github.com/kubewarden/kwctl/issues/418
Depends on https://github.com/kubewarden/policy-evaluator/pull/246 

feat: Add `kwctl scaffold artifacthub` command 
feat: Add `--usage` flag to `kwctl annotate`
Depend on `time` crate.

Example:
```
kwctl scaffold artifacthub \
  --metadata-path metadata.yml \
  --version 0.2.1 \
  --questions-path questions-ui.yml > artifacthub-pkg.yml
```

## Test

Relevant `annotate` unit test has been added, artifacthub functionality is tested in policy-evaluator, and e2e tests are unnecessary.

### Potential improvement

Right now `--version` is mandatory.

### DIFF
--- a/src/annotate.rs
+++ b/src/annotate.rs
@@ -11,17 +11,18 @@ pub(crate) fn write_annotation(
     destination: PathBuf,
     usage_path: Option<PathBuf>,
 ) -> Result<()> {
-    let usage_content: String;
-    let usage = match usage_path {
-        Some(path) => {
-            usage_content = fs::read_to_string(path)
-                .map_err(|e| anyhow!("Error reading questions file: {}", e))?;
-            Some(usage_content.as_str())
-        }
-        None => None,
-    };
+    let usage = usage_path
+        .map(|path| {
+            fs::read_to_string(path).map_err(|e| anyhow!("Error reading usage file: {}", e))
+        })
+        .transpose()?;
     let backend_detector = BackendDetector::default();
-    let metadata = prepare_metadata(wasm_path.clone(), metadata_path, backend_detector, usage)?;
+    let metadata = prepare_metadata(
+        wasm_path.clone(),
+        metadata_path,
+        backend_detector,
+        usage.as_deref(),
+    )?;
     write_annotated_wasm_file(wasm_path, destination, metadata)
 }
 
@@ -211,8 +212,7 @@ mod tests {
         let file_path = dir.path().join("metadata.yml");
         let mut file = File::create(file_path.clone())?;
 
-        let raw_metadata = format!(
-            r#"
+        let raw_metadata = r#"
         rules:
         - apiGroups: [""]
           apiVersions: ["v1"]
@@ -221,8 +221,7 @@ mod tests {
         mutating: false
         backgroundAudit: true
         executionMode: kubewarden-wapc
-        "#
-        );
+        "#;
 
         write!(file, "{}", raw_metadata)?;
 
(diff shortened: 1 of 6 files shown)


viccuad:
rebased on top of main.

src/annotate.rs
```diff
@@ -2,23 +2,34 @@ use crate::backend::{Backend, BackendDetector};
...
 ) -> Result<()> {
+    let usage_content: String;
+    let usage = match usage_path {
+        Some(path) => {
+            usage_content = fs::read_to_string(path)
+                .map_err(|e| anyhow!("Error reading questions file: {}", e))?;
+            Some(usage_content.as_str())
+        }
+        None => None,
+    };
```
flavio, line 22: I don't like the `usage_content` variable. I think you introduced it because `prepare_metadata` expects `usage` to be `Option<&str>` and passing `&usage_content` would produce a `Option<&String>`.
Hence you had to resort to the "external" (to the block) `usage_content` to make the rust borrow checker happy on line 19.

You can convert a `Option<String>` to `Option<&str`> using the [`Option::as_deref`](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_deref) method.


Aside from that, we can also get rid of the "verbosity" of the `match` statement by using this block

```rust
let usage = usage_path
    .map(|path| {
        fs::read_to_string(path).map_err(|e| anyhow!("Error reading questions file: {}", e))
    })
    .transpose()?;
```

and later do:

```rust
let metadata = prepare_metadata(
    wasm_path.clone(),
    metadata_path,
    backend_detector,
    usage.as_deref(),
)?;
```

src/annotate.rs
```diff
@@ -226,6 +246,48 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_kwctl_usage_is_added_when_annotations_is_none() -> Result<()> {
+        let dir = tempdir()?;
+
+        let file_path = dir.path().join("metadata.yml");
+        let mut file = File::create(file_path.clone())?;
+
+        let raw_metadata = format!(
```
flavio, line 256: I don't think we have to convert that to a string, isn't a simple `&str` good enough for `write!`?

src/scaffold.rs
```diff
@@ -209,6 +213,43 @@ pub(crate) fn verification_config() -> Result<String> {
...
+        .map_err(|e| anyhow!("Error unmarshalling metadata {}", e))?;
+    let questions_content: String;
+    let questions = match questions_path {
+        Some(path) => {
+            questions_content = fs::read_to_string(path)
+                .map_err(|e| anyhow!("Error reading questions file: {}", e))?;
+            Some(questions_content.as_str())
+        }
+        None => None,
+    };
```
flavio, line 241: I also don't like this block of code for the same reasons of the `usage_contents` snippet from the `annotate.rs` file.

I think this can be solved in the same way

flavio (changes requested):

viccuad:
Thanks! Indeed, didn't know about `transpose()`, is quite cleaner. Refactor everything on last commit.

flavio (approved):
LGTM, thanks for the fixes

jvanz (approved):

viccuad:
Rebased consuming released policy-evaluator 0.7.0.

PR #439, opened 2023-03-03
docs: Document GitHub Action setup on forks

viccuad:
## Description

Just some things I noticed when forking kwctl and trying to make a release from the fork.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,3 +10,13 @@
    `git tag -s -a -m "vX.Y.Z" vX.Y.Z`. This will trigger the GH Action for
    release. Wait for it to complete and check that it is created.
 5. If needed, edit the GH release description.
+
+## GitHub Actions
+
+For some workflows, GITHUB_TOKEN needs read and write permissions (e.g: to
+perform cosign signatures); if you have forked the repository, you may need to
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -10,3 +10,13 @@
...
+
+## GitHub Actions
+
+For some workflows, GITHUB_TOKEN needs read and write permissions (e.g: to
+perform cosign signatures); if you have forked the repository, you may need to
+change "settings -> actions -> general -> workflow permissions"  to "Read and
+write permissions".
+
+Also, given how the release and release-drafter workflows work, they need git
+tags present; push the tags from origin to your fork.
```
viccuad, line 22: A different approach would be to change:
https://github.com/kubewarden/kwctl/blob/6b1ecf13abff74d94247ab80fef918d9fc97b34c/.github/workflows/release.yml#L92

to:

```diff
-            core.setFailed("Cannot find latest release")
+            core.setOutput('old_release_tag', 'nonexistent')
+            return
```

But then we always open PRs against helm-charts regardless of valid tags or not I suppose.

flavio: Yeah, I would not change that

flavio (approved):

jvanz (approved):

PR #444, opened 2023-03-09
feat: introduce info command

flavio:
Introduce a new subcommand called "info" that displays system information like:

* kwctl version
* What are the Rego builtins supported
* Location of the policy once downloaded on the system


viccuad (approved):
LGTM

PR #450, opened 2023-03-15
fix: grant access to minikube/k3d

flavio:
kwctl is built using rustls enabled. Unfortunately rustls does not support validating IP addresses yet (see https://github.com/kube-rs/kube/issues/1003).

This commit introduces a workaround to address this limitation. This is required to make sure users can run context aware policies against minikube, k3d and other local development solution that deploy kubernetes in this way.


viccuad (approved):
LGTM!
Needs the clippy warnings fixed now, though.

flavio:
> LGTM! Needs the clippy warnings fixed now, though.

doh... I'll handle that and merge it once everything is green

jvanz (approved):
Besides the red CI jobs, LGTM

PR #455, opened 2023-03-17
feat: --output flag to scaffold artifacthub command.

jvanz:
## Description

Adds a new command line argument for the `scaffold artifacthub` subcommand to allow users to define where the artifacthub file content will be written. The new flag is optional. When not defined the content of the file will be written in the stdout.


flavio (approved):
super neat!

PR #459, opened 2023-03-20
chore(deps): update mdcat to latest version

flavio:
By updating to latest version, `info` messages are now shown to our users.
Because of that, we can address some `TODO` items we had left.


flavio:
BTW, there's no change to the `Cargo.lock` because renovate already updated that (it was a patch release and we were previously requiring any patch release of the `1.1` code stream)

viccuad (approved):

jvanz (approved):

Issue #461, opened 2023-03-21
Warning message when executing command `kwctl annotate`


geeksambhu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using the `kwctl` version  `1.6.0-rc4 ` and executing `kwctl annotate` results in the following warning message.

```bash

2023-03-20T18:52:24.287753Z  WARN walrus::module: failed to parse `name` custom section Invalid name type (at offset 3743230)

```

I was trying to work on this [policy](https://github.com/kubewarden/allow-privilege-escalation-psp-policy), and while running  `make annotated-policy.wasm` target which in turns execute `kwctl annotate` command resulted in above warning

### Expected Behavior

This warning resulting from `walrus` should be fixed.

### Steps To Reproduce

1. Install `kwctl` release candidate `1.6.0-rc4 `.
2. clone this repo: [allow-privilege-escalation-psp-policy](https://github.com/kubewarden/allow-privilege-escalation-psp-policy)
3. Execute `make annotated-policy.wasm`

### Environment

```markdown
- OS: OSX
- Architecture: x86_64
```

viccuad:
I can reproduce here too!

walrus hasn't had a release in 2 years either, so maybe is one of its dependencies changing behavior
this is the unreleased commits of walrus:   https://github.com/rustwasm/walrus/compare/0.19.0...master
This catches my attention:  https://github.com/rustwasm/walrus/pull/209

I wonder if this is because we are annotating incorrectly the wasm module, or walrus should be more permissive.

flavio:
This is a warning shown since a long time. It's not a new regression.

It doesn't cause any issue

PR #462, opened 2023-03-21
fix: hide walrus warning message

flavio:
Walrus prints an innocent warning message while annotating a policy.
This can cause users to be confused.

This commit suppresses all the warning messages produced by walrus, but leaves the error ones untouched.

Fixes https://github.com/kubewarden/kwctl/issues/461


viccuad (approved):
LGTM!

PR #480, opened 2023-04-11
Check minimum Kubewarden version.

### LINKED ISSUE #467: Feature Request: handle policies defining the minimum version of Kubewarden to be used

Some policies might leverage host capabilities that are available only starting from a certain Kubewarden release. The policy would be "loadable" by an older version of Kubewarden, but at runtime its evaluation will not work as expected (errors will be raised by the host, leading to the rejection of the request).

Because of that, we are going to extend the policy metadata to allow authors to define a minimum version of Kubewarden that is required. See https://github.com/kubewarden/policy-evaluator/issues/267

### Solution you'd like

kwctl must be able to leverage this new metadata detail.

* [x] `annotate`: nothing special to be done. Right now the annotate command writes whatever is found inside of the given file. Maybe we could do some validation of the file, to ensure the value used to specify the version is a valid semantic version
* [x] `inspect`: print this new metadata field
* [x] `run` and `bench`: check the current version of the Kubewarden stack against the value required by the user. If the current version is too old abort with an error

About the Kubewarden stack:

* Kubewarden stack is identified by a `<major>.<minor>` version. There's no patch level. Because of that the version specified by the user will be truncated to be `<major>.<minor>.0`
* To determine the version of the Kubewarden stack supported by kwctl we will rely on the version specified inside of the `Cargo.toml`. This is can be determined using one of [these solutions](https://stackoverflow.com/questions/51539473/get-cargo-toml-version-at-runtime)
* We will not implement any kind of fancy semantic version check: we will do a simple `<kwctl version> >= <version required by the policy>`

_No response_

### Anything else?

_No response_

jvanz:
## Description

Updates the kwctl run and bench commands to check if program version is compatible with the minimum Kubewarden version required by the policy.
    
It also add a new field in the inspect command to show the minimum Kubewarden version required to run the policy.

Fix #467 

## Test

```shell
$ kwctl inspect registry://ghcr.io/jvanz/policies/allow-privilege-escalation-psp:latest | grep minimum
2023-04-11T18:36:39.193625Z  INFO sigstore::cosign::client_builder: Rekor public key not provided. Rekor integration disabled
2023-04-11T18:36:39.193689Z  INFO sigstore::cosign::client_builder: No Fulcio cert has been provided. Fulcio integration disabled
minimum kubewarden version: 1.7.0 

$ kwctl run registry://ghcr.io/jvanz/policies/allow-privilege-escalation-psp:latest --request-path test_data/req_pod_without_security_context.json
Error: Policy required Kubewarden version 1.7.0 but is running on 1.6.0-rc5

```

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -1,11 +1,16 @@
 use anyhow::{anyhow, Result};
+use lazy_static::lazy_static;
 use policy_evaluator::{
     policy_evaluator::{Evaluator, PolicyExecutionMode},
     policy_evaluator_builder::PolicyEvaluatorBuilder,
     policy_metadata::Metadata,
     ProtocolVersion,
 };
+use semver::{BuildMetadata, Prerelease, Version};
 use std::path::{Path, PathBuf};
+lazy_static! {
+    static ref KUBEWARDEN_VERSION: Version = Version::parse(env!("CARGO_PKG_VERSION")).unwrap();
+}
 
 pub(crate) enum Backend {
     Opa,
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,18 +1,20 @@
+KUBE_API_VERSION?=1.25
+
 .PHONY: build
 build:
-	cargo build --release
+	K8S_OPENAPI_ENABLED_VERSION=$(KUBE_API_VERSION) cargo build --release
```
flavio, line 5: Are you sure this is needed? We had to do that with libraries because we cannot specify the k8s version via the feature flag of k8s-openapi. But with kwctl/policy-server we can do that, and it always worked fine

src/backend.rs
```diff
@@ -120,6 +125,31 @@ impl BackendDetector {
...
+    if let Some(minimum_kubewarden_version) = &metadata.minimum_kubewarden_version {
+        let sanitized_minimum_kubewarden_version = Version {
+            major: minimum_kubewarden_version.major,
+            minor: minimum_kubewarden_version.minor,
+            // Kubewarden stack version ignore patch version number
+            patch: 0,
+            // prerelease and build objects does not implement copy trait.
+            // Thus, we need to clone them
+            pre: minimum_kubewarden_version.pre.clone(),
+            build: minimum_kubewarden_version.build.clone(),
```
flavio, line 140: This has to be fixed like you did with policy-server. Using `empty`. Make sure the comment is dropped

flavio (changes requested):

src/backend.rs
jvanz: Ops... I forgot to change that here. Fixing now...

jvanz:
@flavio, JFYI: I've change the code to avoid a unnecessary `clone()` call.

PR #500, opened 2023-05-20
(feat) added feature to disable colorful output with --no-color flag

### LINKED ISSUE #195: Feature Request: allow disable colorful output

Sometimes, when we want to share the command output (e.g. documentation) the colors chars gets in the way of coping the output to somewhere else. 

### Solution you'd like

I would like to have a flag which disable the colorful output. Something like: `--no-color`

It's possible to remove the color chars using sed:

```console
kwctl inspect registry://ghcr.io/kubewarden/policies/user-group-psp:v0.2.0 | sed -r "s/\x1B\[([0-9]{1,3}(;[0-9]{1,2})?)?[mGK]//g" 
```

### Anything else?

_No response_

adnanhashmi09:
## Description
The `--no-color` flag removes color from the output. It also disables color from the logs genearted by `tracing` crate.

Fixes #195 

## Additional Information
I am modifying the `TERM` env variable to disable color. This is referenced from [here](https://docs.rs/termcolor/1.2.0/src/termcolor/lib.rs.html#255). The changes made to the TERM variable using `env::set_var("TERM", ...)` are specific to the current process and its child processes. Once the process terminates, the environment variable will revert to its previous value.

### DIFF
--- a/src/inspect.rs
+++ b/src/inspect.rs
@@ -18,6 +18,7 @@ use pulldown_cmark_mdcat::{
     terminal::{TerminalProgram, TerminalSize},
 };
 use std::convert::TryFrom;
+use std::env;
 use syntect::parsing::SyntaxSet;
 
 pub(crate) async fn inspect(uri: &str, output: OutputType, sources: Option<Sources>) -> Result<()> {
@@ -235,8 +236,16 @@ impl MetadataPrinter {
         let size = TerminalSize::detect().unwrap_or_default();
         let columns = size.columns;
         let terminal = TerminalProgram::detect();
+        let mut capabilities = terminal.capabilities();
+
+        if let Ok(val) = env::var("TERM") {
+            if val == "dumb" {
+                capabilities.style = None;
+            }
+        }
+
         let settings = pulldown_cmark_mdcat::Settings {
-            terminal_capabilities: terminal.capabilities(),
+            terminal_capabilities: capabilities,
             terminal_size: TerminalSize { columns, ..size },
             syntax_set: &SyntaxSet::load_defaults_newlines(),
             theme: pulldown_cmark_mdcat::Theme::default(),
--- a/src/inspect.rs
+++ b/src/inspect.rs
@@ -13,15 +13,20 @@ use policy_evaluator::{
 };
 use prettytable::{format::FormatBuilder, Table};
 use pulldown_cmark::{Options, Parser};
+use pulldown_cmark_mdcat::TerminalCapabilities;
 use pulldown_cmark_mdcat::{
     resources::NoopResourceHandler,
     terminal::{TerminalProgram, TerminalSize},
 };
 use std::convert::TryFrom;
-use std::env;
 use syntect::parsing::SyntaxSet;
 
-pub(crate) async fn inspect(uri: &str, output: OutputType, sources: Option<Sources>) -> Result<()> {
+pub(crate) async fn inspect(
+    uri: &str,
+    output: OutputType,
+    sources: Option<Sources>,
+    no_color: bool,
+) -> Result<()> {
     let uri = crate::utils::map_path_to_uri(uri)?;
     let wasm_path = crate::utils::wasm_path(uri.as_str())?;
     let metadata_printer = MetadataPrinter::from(&output);
@@ -32,7 +37,7 @@ pub(crate) async fn inspect(uri: &str, output: OutputType, sources: Option<Sourc
     let signatures = fetch_signatures_manifest(uri.as_str(), sources).await;
 
     match metadata {
-        Some(metadata) => metadata_printer.print(&metadata)?,
+        Some(metadata) => metadata_printer.print(&metadata, no_color)?,
         None => return Err(anyhow!(
             "No Kubewarden metadata found inside of '{}'.\nPolicies can be annotated with the `kwctl annotate` command.",
             uri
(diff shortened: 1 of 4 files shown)


viccuad (approved):
LGTM! The implementation is a bit idiosyncratic, but can't think of a better way.

Tested locally, works as expected. An e2e test would be nice for the feature, though.

Many thanks for this contribution!

src/inspect.rs
```diff
@@ -235,8 +236,16 @@ impl MetadataPrinter {
...
         let columns = size.columns;
         let terminal = TerminalProgram::detect();
+        let mut capabilities = terminal.capabilities();
+
+        if let Ok(val) = env::var("TERM") {
+            if val == "dumb" {
+                capabilities.style = None;
+            }
+        }
+
```
flavio, line 246: I don't like to rely on this env variable, which is - from what I understand - a way to communicate between the `main` and the `inspect` functions.

Can't we add a new parameter to the `inspect` function, a `disable_color: bool`?

Then we could do something like that:

```rust
let capabilities = if false {
    TerminalCapabilities {
        style: None,
        ..terminal.capabilities()
    }
} else {
    terminal.capabilities()
};
```

flavio (changes requested):

src/inspect.rs
adnanhashmi09: We could add a new parameter to the inspect function but then we would have to change the signature of a few more functions. `print`, `print_metadata_rules`, `print_metadata_context_aware_resources` and then pass it down to `render_markdown`. Would you like me to implement this?
```rust
impl MetadataPrinter {
    fn print(&self, metadata: &Metadata) -> Result<()> {
        match self {
            MetadataPrinter::Yaml => {
                let metadata_yaml = serde_yaml::to_string(metadata)?;
                println!("{metadata_yaml}");
                Ok(())
            }
            MetadataPrinter::Pretty => {
                self.print_metadata_generic_info(metadata)?;
                println!();
                self.print_metadata_rules(metadata)?;
                println!();
                if !metadata.context_aware_resources.is_empty() {
                    self.print_metadata_context_aware_resources(metadata)?;
                    println!();
                }
                self.print_metadata_usage(metadata)
            }
        }
    }
}
```

flavio: Sure, go ahead with that

kravciak:
> @viccuad I have never written e2e tests in bats. I will look into it and write a test soon. Thanks for the feedback.

I use bash function that removes colors from output, maybe it helps you in the test:
```bash
kwctl load &| sed -r 's:\x1b\[[0-9;]*[mK]::g'
```

adnanhashmi09:
> > @viccuad I have never written e2e tests in bats. I will look into it and write a test soon. Thanks for the feedback.
> 
> I use bash function that removes colors from output, maybe it helps you in the test:
> 
> ```shell
> kwctl load &| sed -r 's:\x1b\[[0-9;]*[mK]::g'
> ```

Thank you for the suggestion but I checked if the output contains ANSI escape sequence.

viccuad (approved):
LGTM! Indeed better. Thanks for the e2e test.

src/main.rs
```diff
@@ -82,6 +82,22 @@ lazy_static! {
...
+    let no_color = matches
+        .get_one::<bool>("no-color")
+        .unwrap_or(&false)
+        .to_owned();
+
+    if no_color {
+        env::set_var("TERM", "dumb");
+    } else {
+        env::set_var("TERM", term_color_support);
+    }
```
flavio, line 100: I think this can be dropped now, isn't it?

src/main.rs
```diff
@@ -82,6 +82,22 @@ lazy_static! {
 #[tokio::main]
 async fn main() -> Result<()> {
     let matches = cli::build_cli().get_matches();
+    let mut term_color_support = "dumb".to_string();
+
+    if let Ok(val) = env::var("TERM") {
+        term_color_support = val
+    }
```
flavio, line 89: I think this can be dropped now, isn't it?

flavio (commented):
thanks for the changes and for the tests, I left some minor comments

src/main.rs
adnanhashmi09: Well no. With prettytable the default is to print with no colors. But the row definitions in the code are something like this: 

```rust
table.add_row(row![Fgbl -> "mutating:", metadata.mutating]);
table.add_row(row![Fgbl -> "background audit support:", metadata.background_audit]);
table.add_row(row![Fgbl -> "context aware:", !metadata.context_aware_resources.is_empty()]);
table.add_row(row![Fgbl -> "execution mode:", metadata.execution_mode]);
```

I couldn't find a global method in prettytable that would disable color. So we would have to rewrite the row definitions  like this:
```rust
if no_color {
  table.add_row(row![Fgbl -> "mutating:", metadata.mutating]);
  table.add_row(row![Fgbl -> "background audit support:", metadata.background_audit]);
  table.add_row(row![Fgbl -> "context aware:", !metadata.context_aware_resources.is_empty()]);
  table.add_row(row![Fgbl -> "execution mode:", metadata.execution_mode]);
}else{
  table.add_row(row!["mutating:", metadata.mutating]);
  table.add_row(row!["background audit support:", metadata.background_audit]);
  table.add_row(row!["context aware:", !metadata.context_aware_resources.is_empty()]);
  table.add_row(row!["execution mode:", metadata.execution_mode]);
}
```

This would cause a lot of redundant code.

PR #518, opened 2023-06-19
feat: support WASI policy

flavio:
Support policies that implement the new WASI protocol.

This policy execution mode is not the best one, but it allows the execution of WASM modules that do not leverage waPC. This is the case of modules built by the official Go compiler.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -445,7 +445,7 @@ checksum = "a3e2c3daef883ecc1b5d58c15adae93470a91d425f3532ba1695849656af3fc1"
 [[package]]
 name = "burrego"
 version = "0.3.4"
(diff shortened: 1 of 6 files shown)


flavio:
@kubewarden/maintainers would you be fine tagging a RC of kwctl 1.7.0? This would make possible to annotate the WASI policies inside of GH actions

jvanz (approved):

jvanz:
> @kubewarden/maintainers would you be fine tagging a RC of kwctl 1.7.0? This would make possible to annotate the WASI policies inside of GH actions

Fine by me. This change deserves a minor release.

flavio:
Moving to blocked, waiting for @viccuad to be back so that we can discuss how to proceed with the release.

viccuad (approved):
LGTM! I'm ok with a `kwctl` RC. I Think too that it deserves a minor version bump.

Issue #529, opened 2023-06-16
Make kwctl aware of new severity and category annotations


viccuad:
With the audit feature, we added support for new Policy Resource annotations (not to confuse with Policy metadata.yaml annotations), for policy severity and policy category. These are:
```yaml
metadata:
  annotations:
    # one of critical, high, medium, low, info. Info is used when policy is in monitor mode
    # Empty by default, set by admin
    io.kubewarden.policy.severity: "high" 
    # freefrom string. Example: "PSP", "Sample", "Signing", "Other", "Compliance"
    # Empty by default, set by admin
    io.kubewarden.policy.category: "PSP"
```

### Acceptance criteria

- Make these annotations work with `kwctl annotate` & `kwctl inspect`, so they are part of metadata.yaml as others.
- Make `kwctl scaffold` treat these, possibly setting empty values for them with an explanation comment.

viccuad:
`kwctl annotate` already gobbles the annotations, and annotates the policy correctly. The annotations appear also in `kwctl inspect`.

The missing part then is `kwctl scaffold`, so they are added to the object `metadata.annotations`, if they are present.

PR #532, opened 2023-07-06
feat: `kwctl scaffold` prints severity and category annotations when they are in policy metadata

viccuad:
## Description

Fix https://github.com/kubewarden/kwctl/issues/529
Depends on https://github.com/kubewarden/policy-evaluator/pull/302

## Test

Unit tests.
Performed a manual e2e test, yet I don't think it is necessary to add a specific e2e test for this.

### Potential improvement

Add an enum for severity (`info`, `low`, `medium`, `high`) in policy-evaluator, to be consumed by kwctl here.
This would allow for `kwctl annotate` to error if the severity annotation in metadata.yaml doesn't conform to those values. Hence, a policy author would realize that they are creating an incorrect default for the severity.
But, since the severity can be changed after doing a `kwctl scaffold`, and we expect cluster admins to set their own severity, I don't see much value in doing this leg work. I could be convinced though!

If the severity would be part of the CRD spec as `spec.severity`, we could ensure its correctness when instantiating policies. But I think it is too early to promote the severity inside of the spec and out of the metadata.annotations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,8 @@ itertools = "0.11.0"
 k8s-openapi = { version = "0.18.0", default-features = false, features = ["v1_26"] }
 lazy_static = "1.4.0"
 pulldown-cmark-mdcat = { version = "2.0.3", default-features = false, features = ["regex-fancy"]}
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.10.0" }
+# policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.10.0" }
+policy-evaluator = { git = "https://github.com/viccuad/policy-evaluator", branch = "main" }
 prettytable-rs = "^0.10"
 pulldown-cmark = { version = "0.9.3", default-features = false }
 regex = "1"
--- a/src/scaffold.rs
+++ b/src/scaffold.rs
@@ -443,23 +443,22 @@ mod tests {
         };
 
         let obj_metadata = build_objmetadata(scaffold_data);
-        assert!(obj_metadata.annotations.is_some());
         assert_eq!(
             obj_metadata
                 .annotations
                 .as_ref()
-                .unwrap()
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -20,7 +20,8 @@ itertools = "0.11.0"
 k8s-openapi = { version = "0.18.0", default-features = false, features = ["v1_26"] }
 lazy_static = "1.4.0"
 pulldown-cmark-mdcat = { version = "2.0.3", default-features = false, features = ["regex-fancy"]}
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.10.0" }
+# policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.10.0" }
+policy-evaluator = { git = "https://github.com/viccuad/policy-evaluator", branch = "main" }
```
jvanz, line 24: I've merged the `policy-evaluator` PR. ;-)

src/scaffold.rs
```diff
@@ -364,6 +416,54 @@ mod tests {
...
+        );
+        assert_eq!(
+            obj_metadata
+                .annotations
+                .as_ref()
+                .unwrap()
+                .get(KUBEWARDEN_ANNOTATION_POLICY_CATEGORY)
+                .unwrap(),
+            &String::from("PSP")
+        );
```
jvanz, line 464: Maybe you can use the [`expect`](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect) method. Thus, you can validate the content and panic with a proper error message when it's invalid.

Cargo.toml
viccuad: thanks!

src/scaffold.rs
viccuad: refactored in https://github.com/kubewarden/kwctl/pull/532. I wonder if this is what you meant, or you meant substituting the whole `assert_eq!()`?

jvanz (approved):

PR #587, opened 2023-08-31
feat: support sha prefix as argument

fabriziosestito:
## Description
This PR adds the support for referring to a policy in the store using a sha value to the following commands:

- run
- inspect
- scaffold manifest
- bench
- push
- rm
- save

Closes: #97 

This feature emulates the docker cli, by accepting a partial sha or a URI.
Since we deal with relative paths, the sha path is used only if a file with the same name does not exist (aka files have precedence). 
If by using a sha prefix multiple policies are found, an error is returned.

E.g.:

```
$ kwctl policies

+--------------------------------------------------------------+----------+---------------+--------------+-----------+
| Policy                                                       | Mutating | Context aware | SHA-256      | Size      |
+--------------------------------------------------------------+----------+---------------+--------------+-----------+
| registry://ghcr.io/kubewarden/policies/pod-privileged:latest | no       | no            | 092f8bc14e4e | 1.36 MB   |
| registry://ghcr.io/kubewarden/policies/safe-labels:latest    | no       | no            | 6a299d04ad3c | 648.47 kB |
| registry://ghcr.io/kubewarden/policies/safe-annotations        | no       | no            | 0c6d3640556 | 2.52 MB   |
+--------------------------------------------------------------+----------+---------------+--------------+-----------+

$ kwctl inspect 0

Error: Multiple policies found with the same prefix: 0

$ kwctl inspect 09

Details                        
title:                        pod-privileged
...
```

## Test

-

### Tradeoff
The test: `test_map_path_to_uri_missing_scheme` in the `utils` mod was removed since it is not relevant anymore, and the function is now depending on the file system.
Anyway, since we agreed to refactor the policy fetcher, this code will change and it will probably moved to the policy fetcher itself.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ lazy_static = "1.4.0"
 pulldown-cmark-mdcat = { version = "2.0.3", default-features = false, features = [
         "regex-fancy",
 ] }
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.11.3" }
+policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.11.4" }
 prettytable-rs = "^0.10"
 pulldown-cmark = { version = "0.9.3", default-features = false }
 regex = "1"
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -6,22 +6,36 @@ use serde_json::json;
 use std::{env, path::PathBuf};
 use url::Url;
 
-pub(crate) fn map_path_to_uri(uri: &str) -> Result<String> {
+pub(crate) fn map_path_to_uri(uri_or_sha_prefix: &str) -> Result<String> {
     let uri_has_schema = Regex::new(r"^\w+://").unwrap();
-    if uri_has_schema.is_match(uri) {
-        return Ok(String::from(uri));
+    if uri_has_schema.is_match(uri_or_sha_prefix) {
+        return Ok(String::from(uri_or_sha_prefix));
     }
-    if PathBuf::from(uri).is_absolute() {
-        Ok(format!("file://{uri}"))
+
+    let path = PathBuf::from(uri_or_sha_prefix);
+    if path.exists() {
+        if path.is_absolute() {
+            Ok(format!("file://{uri_or_sha_prefix}"))
+        } else {
+            Ok(format!(
+                "file://{}/{}",
+                env::current_dir()?
+                    .into_os_string()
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -50,9 +50,12 @@ semver = { version = "1.0.18", features = ["serde"] }
...
-        "rustls-tls-native-roots",
+  "rustls-tls-native-roots",
 ] }
 
 [dev-dependencies]
 rstest = "0.18.1"
 tempfile = "3.7.1"
+
+[patch."https://github.com/kubewarden/policy-fetcher"]
+policy-fetcher = { git = "https://github.com/fabriziosestito/policy-fetcher", branch = "feat/add_store_policy_getters" }
```
flavio, line 61: Reminder: this should be changed before the merge is done

src/utils.rs
```diff
@@ -6,22 +6,36 @@ use serde_json::json;
...
-    if PathBuf::from(uri).is_absolute() {
-        Ok(format!("file://{uri}"))
+
+    let path = PathBuf::from(uri_or_sha_prefix);
+    if path.exists() {
+        if path.is_absolute() {
+            Ok(format!("file://{uri_or_sha_prefix}"))
+        } else {
+            Ok(format!(
+                "file://{}/{}",
```
flavio, line 21: I fear this would break on windows, because of the hardcoded `/` separator. It would probably be better to use [`std::path::Path::join`](https://doc.rust-lang.org/std/path/struct.Path.html#method.join) to build the full path

flavio (commented):
LGTM, I left some minor comments

viccuad (approved):
LGTM. Should we amend some e2e tests to use the sha? (edit: not a blocker, just a musing :)).

flavio (commented):
I like the last round of changes, I'm waiting for the new version of policy-fetcher to be tagged. Then we can merge this PR

fabriziosestito:
> LGTM. Should we amend some e2e tests to use the sha? (edit: not a blocker, just a musing :)).

done in https://github.com/kubewarden/kwctl/pull/587/commits/6b6e4c2a2ef55ce140b6448cf730e5a16743a83c

viccuad (approved):
Thanks, LGTM!
Waiting for merging until policy-fetcher is consumed.

flavio (approved):
LGTM

jvanz (approved):

viccuad (approved):
LGTM, consumed policy-fetcher

PR #594, opened 2023-09-08
test(poc): use rust to test the cli

fabriziosestito:
## Description

This is a POC on how e2e tests could be rewritten in rust, superseeding bats.

The idea is taken from this  book:
https://rust-cli.github.io/book/tutorial/testing.html

Also, some reference in the `ripgrep` project (which uses the same pattern but has a custom macro):
https://github.com/BurntSushi/ripgrep/tree/d2a409f89f1eb8e3ad3990aeb1fff944289534f2/tests

Two dev dependencies were added, `assert_cmd` as a helper to run asserts against the output of the `kwctl` cli programmatically, and `predicates` which helps build composable predicates for the asserts.

Note:
I've modified the Makefile so that `--bins` test is passed to the unit test target. and `--test '*'` is passed to the e2e test target. This way it's possible to separate unit and e2e tests.

## Additional Information
The code could be dried up in several ways, including the usage of `rstest` and macros in general, in particular for the setup functions.

### DIFF
--- a/tests/cli.rs
+++ b/tests/cli.rs
@@ -1,0 +1,68 @@
+use common::{load_fixtures, setup_command};
+use predicates::{prelude::*, str::contains};
+use tempfile::tempdir;
+
+mod common;
+
+const POLICIES: &[&str] = &[
+    // SHA: 59e34f482b40
+    "registry://ghcr.io/kubewarden/tests/pod-privileged:v0.1.9",
+    // SHA: 828617a7cf3e
+    "registry://ghcr.io/kubewarden/tests/safe-labels:v0.1.13",
+];
+
+#[test]
+fn test_policies_empty() {
+    let tempdir = tempdir().unwrap();
+
+    let mut cmd = setup_command(tempdir.path());
+    cmd.arg("policies");
+    cmd.assert().success();
+    cmd.assert().stdout("");
+}
+
+#[test]
+fn test_policies() {
+    let tempdir = tempdir().unwrap();
+    load_fixtures(tempdir.path(), POLICIES);
+
+    let mut cmd = setup_command(tempdir.path());
+    cmd.arg("policies");
+    cmd.assert().success();
+    cmd.assert()
+        .stdout(contains("pod-privileged"))
+        .stdout(contains("v0.1.9"))
+        .stdout(contains("safe-labels"))
+        .stdout(contains("v0.1.13"));
+}
+
+#[test]
+fn test_rm() {
+    let tempdir = tempdir().unwrap();
+    load_fixtures(tempdir.path(), POLICIES);
+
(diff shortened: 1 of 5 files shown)


jvanz (commented):
LGTM. As I'm not big fan of bats, getting rid of it sounds good to me

tests/cli.rs
```diff
@@ -0,0 +1,68 @@
...
+    cmd.assert().stdout(contains("pod-privileged").not());
+}
+
+#[test]
+fn test_rm_with_sha() {
+    let tempdir = tempdir().unwrap();
+    load_fixtures(tempdir.path(), POLICIES);
+
+    let mut cmd = setup_command(tempdir.path());
+    cmd.arg("rm").arg("59e3");
```
flavio, line 61: nitpicking, maybe it would be better to store the policy name + shasum inside of a constant. This would avoid the usage of hard coded checksums here

flavio (approved):
I like this approach, it seems better than the usage of bats

flavio:
waiting for @viccuad's review. We would like to know what he thinks about that, so that we can decide how to move forward with this proposal

jvanz (approved):
Fixing my status review. I've approved but not set the status to "approved" as it should be.

viccuad (approved):
LGTM! indeed is more maintainable and reproducible :).

I had a couple of concerns about the framework, but not the code. I find them satisfied:
-  I would like to run `kwctl` e2e tests as part of github.com/kubewarden/kubewarden-end-to-end-tests. Still doable, we just depend on Rust.
- Would like to remain accessible for contributors writing e2e tests: rethinking this, I feel that community contributors will write more tests in their PRs, if we use Rust.
- I would love to consume testcases from [Dhall](https://dhall-lang.org/) or [Nickel](https://nickel-lang.org/). It seems that this approach makes it more possible.

flavio:
> I would like to run kwctl e2e tests as part of github.com/kubewarden/kubewarden-end-to-end-tests. Still doable, we just depend on Rust.

In this case, I think it would be easier/better to write the tests using bats. Because this is what is being used by the e2e tests. However, I think e2e tests pertaining kwctl should be defined only inside of this repo.

>  I would love to consume testcases from [Dhall](https://dhall-lang.org/) or [Nickel](https://nickel-lang.org/). It seems that this approach makes it more possible.

What do you have in mind? Are you referring to something like cucumber to do BDD testing?

viccuad:
> What do you have in mind? Are you referring to something like cucumber to do BDD testing?

Nothing new, what we have floated about defining testscases and docs in Dhall, to reuse them in several e2e tests, docs, etc.

fabriziosestito:
superseded by: #612

fabriziosestito:
superseded by: https://github.com/kubewarden/kwctl/pull/612

PR #601, opened 2023-09-14
fix: missing step to trigger helm chart update.

jvanz:
## Description

The release CI script is missing a step to trigger the helm update. This is necessary to sync the release among all the major Kubewarden components.

Related to https://github.com/kubewarden/kubewarden-controller/issues/519


viccuad (approved):
LGTM! thanks

PR #630, opened 2023-10-24
feat: add --raw flag to run and bench

fabriziosestito:
## Description

Adds a `--raw` flag to run and bench.
The `--raw` flag specifies that the policy will be evaluated as a raw policy. 

## Test

Updates existing tests.
Adds --raw flag e2e test and fixes.
Updates `tests/data/context-aware-policy-request-pod-creation-all-labels.json` fixture since now we try to unmarshal requests to types and the fixture had missing fields.

## Additional Information

-

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ lazy_static = "1.4.0"
 pulldown-cmark-mdcat = { version = "2.1.0", default-features = false, features = [
   "regex-fancy",
 ] }
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.11.5" }
+policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.12.0" }
 prettytable-rs = "^0.10"
 pulldown-cmark = { version = "0.9.3", default-features = false }
 regex = "1"
--- a/src/run.rs
+++ b/src/run.rs
@@ -1,4 +1,5 @@
 use anyhow::{anyhow, Result};
+use policy_evaluator::admission_request::AdmissionRequest;
 use policy_evaluator::kube;
 use policy_evaluator::{
     constants::*,
@@ -36,6 +37,7 @@ pub(crate) struct PullAndRunSettings {
     pub user_execution_mode: Option<PolicyExecutionMode>,
     pub sources: Option<Sources>,
     pub request: String,
+    pub raw: bool,
     pub settings: Option<String>,
     pub verified_manifest_digest: Option<String>,
     pub fulcio_and_rekor_data: Option<FulcioAndRekorData>,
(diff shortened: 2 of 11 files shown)


src/run.rs
```diff
@@ -118,23 +120,30 @@ pub(crate) async fn prepare_run_env(cfg: &PullAndRunSettings) -> Result<RunEnv>
...
+                }
             }
-        }
-        _ => Err(anyhow!("request to evaluate is invalid")),
-    }?;
+            _ => Err(anyhow!("request to evaluate is invalid")),
+        }?;
+        let adm_req: AdmissionRequest = serde_json::from_value(req_obj)?;
+        ValidateRequest::AdmissionRequest(adm_req)
+    };
```
fabriziosestito, line 142: I'm not sure about this. in the policy server, we expect the raw request to be bewrapped in something like:
```json
{
"request": ...
}
```

The command accept both AdmissionReviews and the inner AdmissionRequest.
It would be nice to implement the same behavior for Raw requests but I am afraid ti could cause confusion if the users sends payloads such as:

```
{
  "request": ""
  "foo": "bar"
}
```
where the `request` key is part of the request payload itself.

Cargo.toml
```diff
@@ -59,3 +59,6 @@ tempfile = "3.8.0"
 assert_cmd = "2.0.12"
 predicates = "3.0.4"
 testcontainers = "0.15"
+
+[patch."https://github.com/kubewarden/policy-evaluator"]
+policy-evaluator = { git = "https://github.com/fabriziosestito/policy-evaluator", branch = "feat/raw-validations" }
```
flavio, line 64: reminder: consume the right version of the crate

src/run.rs
flavio: I think the code you currently wrote is fine. We just have to be explicit inside of the docs about how to use kwctl to process raw requests

tests/e2e.rs
```diff
@@ -9,8 +9,8 @@ use testcontainers::{clients, core::WaitFor};
 mod common;
 
 const POLICIES: &[&str] = &[
-    // SHA: 59e34f482b40
-    "registry://ghcr.io/kubewarden/tests/pod-privileged:v0.1.9",
+    // SHA: 01690a10f9c3
+    "registry://ghcr.io/kubewarden/tests/pod-privileged:v0.2.5",
```
flavio, line 13: oh yes, that was the old policy written with AssemblyScript that cannot be ran anymore.

tests/e2e.rs
```diff
@@ -158,12 +158,28 @@ fn test_run_remote() {
...
+#[test]
+fn test_run_raw() {
+    let tempdir = tempdir().unwrap();
+
+    let mut cmd = setup_command(tempdir.path());
+    cmd.arg("run")
+        .arg("--raw")
+        .arg("--request-path")
+        .arg(test_data("raw.json"))
+        .arg("registry://ghcr.io/fabriziosestito/tests/raw-mutation-policy");
```
flavio, line 176: TODO: update registry

flavio (approved):
LGTM, we have to wait for the test policies to be hosted under the kubewarden org

viccuad (approved):
LGTM! :). Thanks for bumping the test fixtures.

PR #632, opened 2023-10-26
feat: use policyType metadata

fabriziosestito:
## Description

- show policyType when inspecting a policy
- run a policy in raw "mode" if the policyType metadata is set to raw, otherwise the user needs to pass a `--raw` flag (which takes precedence over the metadata, in case the policy is not annotated)
## Test
Adds e2e tests
## Additional Information
Bumps Cargo toml to `v1.9.0-rc1`

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -1,6 +1,7 @@
 use anyhow::{anyhow, Result};
 use policy_evaluator::admission_request::AdmissionRequest;
 use policy_evaluator::kube;
+use policy_evaluator::policy_metadata::PolicyType;
 use policy_evaluator::{
     constants::*,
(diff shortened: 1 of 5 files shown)


flavio (approved):

src/run.rs
```diff
@@ -1,6 +1,7 @@
 use anyhow::{anyhow, Result};
 use policy_evaluator::admission_request::AdmissionRequest;
 use policy_evaluator::kube;
+use policy_evaluator::policy_metadata::PolicyType;
```
viccuad, line 4: quibble: it would be cleaner in line 11

viccuad (approved):

PR #671, opened 2023-12-21
Upgrade artifact upload/download to v4

kravciak:
## Description

Fix #XXX

## Test


viccuad (approved):
I'm aproving, but I have a slight doubt that the download keeps the same folders, and that the following "Upload release assets" is going to find the needed files.

kravciak:
> I'm aproving, but I have a slight doubt that the download keeps the same folders, and that the following "Upload release assets" is going to find the needed files.

You might be right, I will double check & test the changes - thanks!

PR #681, opened 2024-01-09
ci: Fix release job by uploading artifacts only once

viccuad:
## Description

If not, both the workflow run of the job `build-linux-binaries` for `aarch64` and `x86_64` try to upload the same file, which results in error.

See:  https://github.com/kubewarden/kwctl/actions/runs/7460505456/job/20298712321

## Test

Untested.


flavio (approved):
Thanks for the fix

PR #698, opened 2024-01-25
build: Bump version to v1.10.0

viccuad:
## Test

CI


flavio (approved):
This is going to have the final 1.10.0 release to consume latest versions of k8s-openapi and kube.

I think this is fine, everything tested green inside of the unit, integration and e2e tests

PR #723, opened 2024-02-16
ci: Add codecov support

viccuad:
## Description

Fix https://github.com/kubewarden/kwctl/issues/722

## Test

CI.
We should get the codecov message.


viccuad:
ready for review, now tests are correctly separated.

flavio (approved):

PR #765, opened 2024-04-03
test: rewrite airgap tests as integration tests

### LINKED ISSUE #706: kwctl: allow e2e tests to be run locally
### Problem

Airgap e2e tests are scripted [by using github actions](https://github.com/kubewarden/kwctl/blob/main/.github/workflows/ci.yml#L115).

This makes impossible to run these tests locally.

### Admission criteria

* [x] e2e tests can be run locally, with a Makefile target

fabriziosestito:
## Description

Fix #706 by rewriting the airgap e2e tests in Rust.
Airgrap scripts were modified to allow changing the kwctl command so that we can inject the test binary. 
An environment variable is used as we don't want to expose this to the end-user.


.github/workflows/ci.yml
```diff
@@ -143,38 +143,3 @@ jobs:
...
-      - name: Verify policies in local registry
-        run: |
-          kwctl pull registry://localhost:5000/kubewarden/tests/pod-privileged:v0.1.9 \
-              --sources-path tests/data/airgap/insecure.yml
-          kwctl pull registry://localhost:5000/kubewarden/pod-privileged-policy/releases/download/v0.1.6/policy.wasm \
-              --sources-path tests/data/airgap/insecure.yml
-      - name: Clean up - delete registry
-        if: always()
-        run: |
-          docker rm -f ${{ env.CONTAINER_ID }}
```
jvanz, line 180: Cannot we just call the new test here? As far as I can remember, we add this test because the scripts failed due some regression just because we forgot to update them. I would like to avoid user to be the ones whom find this issues. Maybe just set the job as optional in the PR checks.

fabriziosestito: @jvanz the test is running as a part of the e2e tests now, no need to add a new job for it

flavio (approved):
LGTM

PR #777, opened 2024-04-15
ci: Build release binaries for `main`, `feat-*` branches

viccuad:
## Description

Build release binaries for `main`, `feat-*` branches.

I believe this is the correct way of triggering these kind of builds for main. The release workflow doesn't do anything for builds, just depends and workflow_calls the build workflow.

The resulting binaries are available to download from the job run.

## Test

See the run for main on my fork: https://github.com/viccuad/kwctl/actions/runs/8690166643/job/23829437135

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,11 @@
 name: kwctl build
 on:
   workflow_call:
+  push:
+    branches:
+      - "main"
+      - "feat-**"
+
 env:
   CARGO_TERM_COLOR: always
 


.github/workflows/build.yml
```diff
@@ -1,6 +1,11 @@
 name: kwctl build
 on:
   workflow_call:
+  push:
+    branches:
+      - "main"
+      - "feat-**"
```
jvanz, line 7: Is should not be `feat-*` (single `*`)?

jvanz (approved):

PR #834, opened 2024-06-19
feat: scaffold KW policy from Kubernetes ValidatingAdmissionPolicy

flavio:
Add the new `scaffold vap` subcommand. This allows to create a Kubewarden ClusterAdmissionPolicy starting from a Kubernetes ValidatingAdmissionPolicy.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -465,6 +465,27 @@ fn subcommand_scaffold() -> Command {
             .help("Policy URI or SHA prefix. Supported schemes: registry://, https://, file://. If schema is omitted, file:// is assumed, rooted on the current directory."),
     );
 
+    let mut vap_args = vec![
+        Arg::new("cel-policy")
+            .long("cel-policy")
+            .value_name("URI")
+            .default_value("ghcr.io/kubewarden/policies/cel-policy:latest")
+            .help("The CEL policy module to use"),
+        Arg::new("policy")
+            .long("policy")
+            .short('p')
+            .required(true)
+            .value_name("VALIDATING-ADMISSION-POLICY.yaml")
+            .help("The file containining the ValidatingAdmissionPolicy definition"),
+        Arg::new("binding")
+            .long("binding")
+            .short('b')
+            .required(true)
+            .value_name("VALIDATING-ADMISSION-POLICY-BINDING.yaml")
+            .help("The file containining the ValidatingAdmissionPolicyBinding definition"),
+    ];
+    vap_args.sort_by(|a, b| a.get_id().cmp(b.get_id()));
+
     let mut subcommands = vec![
         Command::new("verification-config")
             .about("Output a default Sigstore verification configuration file"),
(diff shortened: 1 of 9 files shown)


src/scaffold.rs
```diff
@@ -328,9 +338,171 @@ pub(crate) fn artifacthub(
...
+    Ok(cluster_admission_policy)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
+    use rstest::*;
+
+    const CEL_POLICY_MODULE: &str = "ghcr.io/kubewarden/policies/cel-policy:latest";
+    const VAP_WITHOUT_VARIABLES: &str = r#"
```
fabriziosestito, line 449: we could use fixtures here instead (`tests/data` folder)

src/scaffold.rs
```diff
@@ -328,9 +338,171 @@ pub(crate) fn artifacthub(
...
+        ));
+    }
+
+    let mut settings = serde_yaml::Mapping::new();
+
+    // migrate CEL variables
+    if let Some(vap_variables) = vap_spec.variables {
+        let vap_variables: Vec<serde_yaml::Value> = vap_variables
+            .iter()
+            .map(|v| serde_yaml::to_value(v).expect("cannot convert VAP variable to YaML"))
```
fabriziosestito, line 387: typo: YAML

src/scaffold.rs
```diff
@@ -328,9 +338,171 @@ pub(crate) fn artifacthub(
...
+            .collect();
+        settings.insert("variables".into(), vap_variables.into());
+    }
+
+    // migrate CEL validations
+    if let Some(vap_validations) = vap_spec.validations {
+        let kw_cel_validations: Vec<serde_yaml::Value> = vap_validations
+            .iter()
+            .map(|v| serde_yaml::to_value(v).expect("cannot convert VAP validation to YaML"))
```
fabriziosestito, line 396: typo: YAML

jvanz (approved):

fabriziosestito (changes requested):
Overall LGTM, a few nitpicks here and there.
However, we need to add a warning when no version is set, informing the user that we are defaulting to the latest tag.

fabriziosestito (approved):
@flavio approved modulo the warning message

src/cli.rs
```diff
@@ -465,6 +465,27 @@ fn subcommand_scaffold() -> Command {
             .help("Policy URI or SHA prefix. Supported schemes: registry://, https://, file://. If schema is omitted, file:// is assumed, rooted on the current directory."),
     );
 
+    let mut vap_args = vec![
+        Arg::new("cel-policy")
+            .long("cel-policy")
+            .value_name("URI")
+            .default_value("ghcr.io/kubewarden/policies/cel-policy:latest")
```
viccuad, line 472: I would be ok with "ghcr.io/kubewarden/policies/cel-policy:<version>" so by default the user is forced to edit and doesn't instantiate a possible broken policy.

viccuad (approved):
LGTM.

flavio:
I've just pushed a commit that introduces the check about `latest` image being used.

When the policy being referenced is `<anything>:latest` or `<anything>` (no tag -> `latest`), the tool will print a warning. The warning goes to the STDERR, which means the stdout containing the YAML file is not disrupted

flavio:
BTW, I'll squash the commits together once you're done with the review

fabriziosestito:
:shipit:

PR #880, opened 2024-08-07
feat: introduce scaffold admission-request command

flavio:
This command creates a Kubernetes AdmissionRequest object starting from a Kubernetes Resource definition.

The following command:

```console
kwctl scaffold \
  admission-request \
  --operation CREATE \
  --object ingress.yaml
```

produces an AdmissionRequest CREATE event for the given Ingress object.

Currently, only the CREATE operation is supported. The other operations will be added in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Built binaries for `Linux x86_64`, `Windows x86_64`, `MacOS x86_64` and `MacOS
 aarch64 (M1)` are available in [GH Releases](https://github.com/kubewarden/kwctl/releases).
 
 There is also:
+
 - Community-created [Homebrew 🍺 formula for kwctl](https://formulae.brew.sh/formula/kwctl)
 - Community-created [AUR 🐧 package](https://aur.archlinux.org/packages/kwctl-bin)
 
@@ -75,9 +76,9 @@ Policies can be downloaded using the `pull` command.
 The name of the policy must be expressed as a url with one of the
 following protocols:
 
-* `http://`: pull from a HTTP server
-* `https://`: pull from a HTTPS server
-* `registry://`: pull from an OCI registry
+- `http://`: pull from a HTTP server
+- `https://`: pull from a HTTPS server
+- `registry://`: pull from an OCI registry
 
 Pulling from a registry, by tag:
 
@@ -91,6 +92,7 @@ same way as with regular container images):
 ```console
 kwctl pull registry://ghcr.io/kubewarden/policies/psp-capabilities@sha256:61ef63621fa5be8e422881d96d05edfef810992fbf9468e35d1fa5ae815bd97c
 ```
+
 Note well, the shasum is the digest of the OCI artifact containig the policy.
 This value can be obtained using a tool like [crane](https://github.com/google/go-containerregistry/blob/main/cmd/crane/README.md):
 
@@ -100,11 +102,10 @@ crane digest ghcr.io/kubewarden/policies/psp-capabilities:v0.1.6
 
 ### Run a policy locally
 
-`kwctl` can be used to run a policy locally, outside of Kubernetes. This can be used 
+`kwctl` can be used to run a policy locally, outside of Kubernetes. This can be used
 to quickly evaluate a policy and find the right settings for it.
 
 The evalution is done against a pre-recorded [`AdmissionReview`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#request).
-
 
 Running a policy locally:
 
@@ -117,6 +118,19 @@ kwctl run \
 
 Policy configuration can be passed on the CLI via the `--settings-json` flag
 or can be loaded from the disk via the `--settings-path` flag.
+
+#### Scaffold AdmissionReview from a Kubernetes resource
+
+It's possible to scaffold an `AdmissionReview` object from a Kubernetes resource:
+
+```console
+kwctl scaffold \
+  admission-request \
+  --operation CREATE \
+  --object ingress.yaml
+```
+
+The output of the above command can be used by the `run` command.
 
 ### Annotate a policy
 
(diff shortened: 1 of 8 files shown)


flavio:
Depends on https://github.com/kubewarden/policy-evaluator/pull/548

README.md
```diff
@@ -299,16 +312,16 @@ Then start a new shell or run `source ~/.zshrc` once.
...
+When you download a [kwctl release](https://github.com/kubewarden/kwctl/releases/) each zip file contains two
 files that can be used for verification: `kwctl.sig` and `kwctl.pem`.
 
 In order to verify kwctl you need cosign installed, and then execute the following command:
 
 ```
 cosign verify-blob \
   --signature kwctl-linux-x86_64.sig \
-  --cert kwctl-linux-x86_64.pem kwctl-linux-x86_64 
+  --cert kwctl-linux-x86_64.pem kwctl-linux-x86_64
```
viccuad, line 324:
```suggestion
  --cert kwctl-linux-x86_64.pem kwctl-linux-x86_64 \
```

src/cli.rs
```diff
@@ -486,6 +486,25 @@ fn subcommand_scaffold() -> Command {
...
+            .short('o')
+            .required(true)
+            .value_name("TYPE")
+            .value_parser(PossibleValuesParser::new(["CREATE"])) //TODO: add UPDATE and DELETE
+            .help("Kubewarden Custom Resource type"),
+        Arg::new("object")
+            .long("object")
+            .value_name("PATH")
+            .help("The file containing the new object being admitted"),
+        Arg::new("old-object")
```
viccuad, line 501: maybe worth commenting until we have UPDATE.

src/scaffold/admission_request.rs
```diff
@@ -0,0 +1,801 @@
...
+    pub static ref DEFAULT_ROOT: ProjectDirs =
+        ProjectDirs::from("io.kubewarden", "", "kubewarden").unwrap();
+    pub static ref DEFAULT_KWCTL_CACHE: PathBuf = DEFAULT_ROOT.cache_dir().join("kwctl");
+    pub static ref RESOURCE_CATALOG_FILE: PathBuf =
+        DEFAULT_KWCTL_CACHE.join("resource_catalog.json");
+}
+
+const FALLBACK_API_RESOURCE_PLURAL_NAME: &str = "this-is-the-plural-name-of-the-resource-this-information-is-not-used-by-policies-and-requires-a-connection-to-an-api-server-to-be-obtained";
+
+/// Types of AdmissionRequest operation we can scaffolf
```
viccuad, line 35:
```suggestion
/// Types of AdmissionRequest operation we can scaffold
```

src/scaffold/admission_request.rs
```diff
@@ -0,0 +1,801 @@
...
+    object: Option<PathBuf>,
+    old_object: Option<PathBuf>,
+) -> Result<()> {
+    validate_params(&operation, object.as_ref(), old_object.as_ref())?;
+
+    let output = match operation {
+        Operation::Create => {
+            scaffold_create(
+                RESOURCE_CATALOG_FILE.to_path_buf(),
+                kube_client,
```
viccuad, line 262: are we using the correct kube_client here? shouldn't we be using the newly created function?

viccuad (commented):
LGTM, just one doubt about kube_client.

src/cli.rs
flavio: I would like to do that with another PR, right after this one gets merged

src/scaffold/admission_request.rs
flavio: We were using the right function, the naming was not making that evident, though.

I've fixed that with ddd8916b938fe778e42e5c8c5712131ef116d28f

viccuad (approved):

jvanz (commented):
LGTM. But I have a question in mind, what happens if the user has multiple contexts in the kubeconfig? Will the catalog contains the kinds from the default context? I'm wondering  about the UX for the user which is building a request for a CRDs which is not in the "default" context. 

If this is an issue at all,  I don't think this is a huge issue as well and we do not need to fix it now. But if it is, is there an easy workaround for it? Like defining the context to be used or fetch resources from all the contexts ( I'm not sure if this is a good idea in end end as well ) .

flavio:
> LGTM. But I have a question in mind, what happens if the user has multiple contexts in the kubeconfig? Will the catalog contains the kinds from the default context? I'm wondering about the UX for the user which is building a request for a CRDs which is not in the "default" context.

The catalog will be built from the default context.

> If this is an issue at all, I don't think this is a huge issue as well and we do not need to fix it now. But if it is, is there an easy workaround for it? Like defining the context to be used or fetch resources from all the contexts ( I'm not sure if this is a good idea in end end as well ) .

Going through all the contexts would take time and also possible timeouts because maybe some of the clusters are not reachable.

It's good that you pointed out this possible limitation, but I would not focus on that right now. The scaffold code always generate something that will work for 99.9% of the policies.

We can find a solution to this corner case situation if someone complains about that.

jvanz (approved):

src/scaffold/admission_request.rs
```diff
@@ -0,0 +1,801 @@
...
+    resources: HashMap<String, APIResource>,
+    #[serde(skip)]
+    restored_from: ApiResourceCatalogRestoredFrom,
+}
+
+impl ApiResourceCatalog {
+    /// Create a new catalog. The function will try to load the catalog from disk, if it fails
+    /// it will build it by querying the Kubernetes API server.
+    /// If querying the API server fails, the function will return an empty catalog.
+    pub async fn new<F, Fut>(resource_catalog_file: PathBuf, build_kubeclient_fn: F) -> Self
```
fabriziosestito, line 89: let's add a comment explaining why we need to defer the creation of the client passing a fn here.

flavio: done

Issue #852, opened 2024-07-04
Fix kwctl coverage and reach close to 70%


viccuad:
### Acceptance criteria

In reality, we have ~65% code coverage already.  Check why codecov coverage is not recognizing the tests.
Add the needed tests to reach ~70% coverage.

flavio:
We need to fix the codecov configuration, I just checked and we have only [13% of code coverage](https://app.codecov.io/gh/kubewarden/kwctl) which is wrong

PR #948, opened 2024-10-22
ci: Increase timeout for coverage job

viccuad:
## Description

This is in hopes of reducing tarpaulin time below 15 mins,
so it fits in GHA timeout.

Right now it times out after 15 mins:
https://github.com/kubewarden/kwctl/actions/runs/11442780601

Relates to https://github.com/kubewarden/kwctl/issues/852.

## Test

Tested locally, but cannot know about GHA timeout.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ coverage: coverage-unit-tests coverage-e2e-tests
 .PHONY: coverage-unit-tests
 coverage-unit-tests:
 	cargo tarpaulin --verbose --skip-clean --engine=llvm \
-		--all-features --bins --follow-exec \
+		--all-features --implicit-test-threads --bins \
 		--out xml --out html --output-dir coverage/unit-tests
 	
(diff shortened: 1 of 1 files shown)


flavio:
I'm fine with that, but I think we also realized we were running tarpaulin with some flags that made it slower than needed.

@fabriziosestito do you remember which project was fixed?

I see policy-server has some different options: https://github.com/kubewarden/policy-server/blob/9737ba8cfa17f578f6e6451f938611c98be8cbd8/Makefile#L46-L49

viccuad:
Good point. I have force-pushed, removing the timeout increase and setting the same args as we use for policy-server instead.

flavio:
It still got killed, maybe we have to raise the limit too, but to something higher than 30 minutes

viccuad:
Note that the CI run takes the code from main, not the PR though.

PR #956, opened 2024-10-23
automation: switch to cargo-llvm-cov

flavio:
Switch to cargo-llvm-cov to collect code coverage results. This is required because tarpaulin is consuming too much memory, leading to our GHA being killed.

See https://github.com/xd009642/tarpaulin/issues/1639 for more details.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,6 @@ jobs:
   coverage:
     name: coverage
     runs-on: ubuntu-latest
-    continue-on-error: true # https://github.com/xd009642/tarpaulin/issues/1639
     steps:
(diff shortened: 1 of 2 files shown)


flavio:
It works, and what a boost! :rocket: :smile:

flavio:
Some thoughts:

- unit tests and e2e tests are run at the same time by `cargo llvm-cov`
- running all the tests + coverage is fast: it took 7 minutes. Maybe we should drop the `unit tests` and the `e2e tests` actions and just run the `coverage` one

What do you think? If you agree with that, should we do these changes with another PR?

viccuad (approved):
Looks great! Thanks for re-establishing the cosign-installer.

I vote for keeping the unit and e2e tests separately on CI, they are shorter and run in parallel. Plus, we test the local make targets that way too.

PR #1030, opened 2025-01-13
fix(ci): chore context for patch updates and change schedule.

jvanz:
## Description

Updates the configuration used to bump patch versions of the dependencies to use the "chore" type in the semantic commits and follow the default schedule for renovate bot ("schedule:nonOfficeHours").

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -29,12 +29,13 @@
         "*"
       ],
       "automerge": true,
-      "schedule": ["* 0-3 * * *"]
+      "semanticCommits": "enabled",
+      "semanticCommitType": "build"
     }
   ],
   "rebaseWhen": "behind-base-branch",
   "prHourlyLimit": 0,
   "prConcurrentLimit": 0,
   "semanticCommits": "enabled",
-  "semanticCommitType": "chore"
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -29,7 +29,8 @@
         "*"
       ],
       "automerge": true,
-      "schedule": ["* 0-3 * * *"]
+      "semanticCommits": "enabled",
+      "semanticCommitType": "chore"
```
jvanz, line 33: This is necessary to properly update the changelog generated by release drafter

fabriziosestito: shouldn't the commit type be `build(deps)` ?

jvanz: I do not know. I'm fine with `build(deps)` as well. I just do not want to use `fix(...)`.

jvanz: Yeah, I think may want to use `build(deps)`. https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

jvanz:
This will be superseded by a change pointing to this [file](https://github.com/kubewarden/github-actions/pull/148)

Issue #1044, opened 2025-01-20
Propagate some metadata.yml annotations into the OCI image manifest on `kwctl push`


viccuad:
Part of https://github.com/kubewarden/kubewarden-controller/issues/839.

### Acceptance criteria

- [ ] Propagate `io.kubewarden.policy.source` from metadata.yml into an OCI image manifest annotation under `org.opencontainers.image.source`.
 This is done by populating the Config.annotations here: https://github.com/viccuad/policy-fetcher/blob/dee2b67a903ad9f0faa7b840d2ce8adc85ade88c/src/registry/mod.rs#L236
- [ ] Add necessary tests.
- [ ] Document this behaviour.

PR #1081, opened 2025-02-12
feat: OCI annotations

flavio:
When pushing a policy to an OCI registry, add all the Kubewarden annotations found inside of policy's metadata to the resulting OCI manifest.

The KW annotations that are multi-line will be skipped, since they are not compliant with the OCI spec.

Moreover, the `io.kubewarden.policy.source` is propaged as `org.opencontainers.image.source` to allow tools like renovatebot to handle policy updates.

Fixes https://github.com/kubewarden/kwctl/issues/1044

Finally, I've added a `taplo.toml` configuration to the project to keep our `Cargo.toml` file nice and tidy.

### DIFF
--- a/src/push.rs
+++ b/src/push.rs
@@ -1,7 +1,15 @@
+use std::{collections::BTreeMap, fs, path::PathBuf};
+
 use anyhow::{anyhow, Result};
-use policy_evaluator::policy_fetcher::{registry::Registry, sources::Sources};
-use policy_evaluator::policy_metadata::Metadata;
-use std::{fs, path::PathBuf};
+use policy_evaluator::{
+    constants::KUBEWARDEN_ANNOTATION_POLICY_SOURCE,
+    policy_fetcher::{
+        oci_client::annotations::ORG_OPENCONTAINERS_IMAGE_SOURCE, registry::Registry,
+        sources::Sources,
+    },
+    policy_metadata::Metadata,
+};
+use tracing::warn;
 
 use crate::backend::BackendDetector;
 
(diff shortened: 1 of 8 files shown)


src/push.rs
```diff
@@ -44,3 +53,109 @@ fn can_be_force_pushed_without_metadata(
...
+/// Augment the annotations with the `org.opencontainers.image.source`
+/// annotation, if the `io.kubewarden.policy.source` annotation is present.
+fn build_oci_annotations(annotations: BTreeMap<String, String>) -> BTreeMap<String, String> {
+    // filter all the multi-line annotations, they are not supported by the OCI spec
+    let mut annotations: BTreeMap<String, String> = annotations
+        .iter()
+        .filter(|(k, v)| {
+            let filter = v.lines().count() <= 1;
+            if filter {
+                info!(
```
fabriziosestito, line 66: maybe warn?

fabriziosestito (approved):

src/push.rs
flavio: done

flavio:
I've also updated the help message to include more detail. This information is part of the output of `kwctl docs`, hence it's going to land on docs.kubewarden.io automatically.

```console
$ kwctl push --help
Pushes a Kubewarden policy to an OCI registry

Usage: kwctl push [OPTIONS] <policy> <uri>

Arguments:
  <policy>
          Policy to push. Can be the path to a local file, a policy URI or the SHA prefix of a policy in the store.

  <uri>
          Policy URI. Supported schemes: registry://

Options:
      --docker-config-json-path <PATH>
          Path to a directory containing the Docker 'config.json' file. Can be used to indicate registry authentication details

  -f, --force <force>
          Push also a policy that is not annotated

  -o, --output <PATH>
          Output format

          [default: text]
          [possible values: text, json]

      --sources-path <PATH>
          YAML file holding source information (https, registry insecure hosts, custom CA's...)

  -h, --help
          Print help (see a summary with '-h')

The annotations found inside of policy's metadata are going to be part of the OCI manifest.
The multi-line annotations are skipped because they are not compatible with the OCI specification.
The 'io.kubewarden.policy.source' annotation is propaged as 'org.opencontainers.image.source' to allow tools like
renovatebot to detect policy updates.

```

viccuad (approved):
LGTM. From the looks of it, the test images have been updated as the tests pass.

flavio:
fixed the last commit to follow conventional commits guidelines

PR #1105, opened 2025-03-04
feat: add gh release to artifacthub scaffold

fabriziosestito:
## Description  

This PR introduces a GitHub release tag to override the version in URL links. See [[kubewarden/policy-evaluator#644](https://github.com/kubewarden/policy-evaluator/pull/644)](https://github.com/kubewarden/policy-evaluator/pull/644).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ k8s-openapi = { version = "0.24.0", default-features = false, features = [
 ] }
 lazy_static = "1.4.0"
 pem = "3"
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.20.1" }
+policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.21.0" }
 prettytable-rs = "^0.10"
 pulldown-cmark = { version = "0.12.1", default-features = false }
 pulldown-cmark-mdcat = { version = "2.7.1", default-features = false }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -21,7 +21,8 @@ k8s-openapi = { version = "0.24.0", default-features = false, features = [
 ] }
 lazy_static = "1.4.0"
 pem = "3"
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.20.1" }
+#policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.20.1" }
+policy-evaluator = { git = "https://github.com/fabriziosestito/policy-evaluator", branch = "feat/add-gh-release-to-artifacthub-generator" }
```
fabriziosestito, line 25: change this

src/cli.rs
```diff
@@ -429,6 +429,13 @@ fn subcommand_scaffold() -> Command {
             .number_of_values(1)
             .value_name("VALUE")
             .help("Semver version of the policy"),
+        Arg::new("gh-release-tag")
```
fabriziosestito, line 432: There are no tests for the artifacthub scaffold command. However, the logic is extensively tested in PolicyEvaluator. Should we add a test here? If so, what aspects should we cover?

flavio (approved):

src/cli.rs
flavio: I have mixed feelings about that... the only things not tested here would be the handling of the flags, reading the contents of the files (like questions.yml). I wonder if this is worth. I would track that as technical debt and deal with that later

viccuad (approved):

flavio (approved):

viccuad (approved):

PR #1116, opened 2025-03-07
feat: add spinner while pulling resources

### LINKED ISSUE #2: Add progress bar when pulling
We should have a nice progress bar that helps understand what is happening behind the scenes. Right now pull doesn't show anything at all

maanugh:
## Description

Add a progress bar to the policy pull operation to provide visual feedback during fetching. This enhancement integrates the indicatif crate to display a spinner-style progress indicator that shows when a policy is being pulled, updates with status messages during the operation, and concludes with appropriate success or failure notifications upon completion, improving the user experience without altering the existing functionality.

Fixes #2

`Spinner `

image

// @flavio

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ wasmparser = "0.226"
 reqwest = { version = "0", default-features = false, features = [
   "rustls-tls-native-roots",
 ] }
+indicatif = "0.17"
 
 [dev-dependencies]
 assert_cmd     = "2.0.14"
--- a/src/pull.rs
+++ b/src/pull.rs
@@ -1,4 +1,7 @@
+use std::time::Duration;
+
 use anyhow::Result;
+use indicatif::{ProgressBar, ProgressStyle};
 use policy_evaluator::policy_fetcher::{
     fetch_policy, policy::Policy, sources::Sources, PullDestination,
 };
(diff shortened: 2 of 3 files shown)


viccuad (approved):
Many thanks for this contribution! LGTM, approving.

jvanz (approved):
That's cool! Thanks!

flavio (approved):
That's cute, I like it! :heart: 

Thanks for the contribution

Cargo.toml
```diff
@@ -53,6 +53,7 @@ wasmparser = "0.226"
 reqwest = { version = "0", default-features = false, features = [
   "rustls-tls-native-roots",
 ] }
+indicatif = "0.17.11"
```
flavio, line 56: We can be more relaxed about the patch version

```suggestion
indicatif = "0.17"
```

src/pull.rs
```diff
@@ -1,14 +1,32 @@
 use anyhow::Result;
+use indicatif::{ProgressBar, ProgressStyle};
 use policy_evaluator::policy_fetcher::{
     fetch_policy, policy::Policy, sources::Sources, PullDestination,
 };
+use std::time::Duration;
```
flavio, line 6: nitpick: can you move the `std` include at the top of the file, then leave one empty line and have the other imports?

```suggestion
use std::time::Duration;

use anyhow::Result;
use indicatif::{ProgressBar, ProgressStyle};
use policy_evaluator::policy_fetcher::{
    fetch_policy, policy::Policy, sources::Sources, PullDestination,
};
```

This is best practice used by Rust projects to handle imports. I acknowledge this is something we are not doing everywhere, this is something we're slowly fixing.

src/pull.rs
```diff
@@ -1,14 +1,32 @@
...
     uri: &str,
     sources: Option<&Sources>,
     destination: PullDestination,
 ) -> Result<Policy> {
-    fetch_policy(uri, destination, sources)
+    let pb = ProgressBar::new_spinner();
+    pb.set_style(
+        ProgressStyle::default_spinner()
+            .template("{spinner:.green} {msg}")
+            .unwrap(),
```
flavio, line 17: nitpick: I prefer to not have `unwrap` inside of prod code. In this case the code should never fail, hence it's fine to use an `expect` instead of raising the error

```suggestion
            .expect("cannot set spinner template"),
```

flavio (changes requested):
Some small changes. Also, please [sign your commit](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), this is one of the requirements we have (see the failure message on the PR)

Cargo.toml
maanugh: fixed

src/pull.rs
maanugh: Ok, I will keep in mind next time.

flavio (approved):

flavio:
@maanugh thanks for your changes!

PR #1122, opened 2025-03-10
feat: scaffold helm chart 

fabriziosestito:
## Description
WIP


fabriziosestito:
Closing since we agreed on using another approach, see: https://github.com/kubewarden/policy-catalog

PR #1127, opened 2025-03-12
refactor!: Change `kwctl inspect` colored output

viccuad:
## Description

Drop `pulldown-cmark`,`pulldown-cmark-mdcat`,`syntect` since `pulldown-cmark-mdcat` is now unmaintained.

Use `termimad` instead.

Before, `kwctl inspect` had syntax coloring for code blocks such as "rules" and those inside the "description".
Now, code blocks don't have syntax coloring.

There's minor variations on printing the "description" markdown. Markdown URL links are flat rendered by `termimad`.

Maintained the feature of disabling color when piping to non-interactive terminal by depending on `is-terminal` (`is-terminal` is already a dependency of `termimad`, yet `termimad` doesn't provide this functionality on its own).

## Test

Locally.

Now:
now

Before:
before

### Tradeoff

Now we print code blocks without syntax coloring.

### Potential improvement

Alternatives:
Fork `pulldown-cmark-mdcat`. To me, `pulldown-cmark`, being a pull parser, provides the only sane approach to bring back the syntax colouring feature. Tried `tui-markdown`, `comrak`, `ratatui + syntect`, `termimad + syntect`, but without the events of pulldown it becomes increasingly difficult to know when one is in a fenced named code block (there can be nested ones).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,8 @@ clap_complete = "4.5"
 directories = "6.0.0"
 flate2 = "1.1"
 humansize = "2.1"
+indicatif = "0.17"
+is-terminal = "0.4.16"
 itertools = "0.14.0"
 k8s-openapi = { version = "0.24.0", default-features = false, features = [
   "v1_30",
@@ -23,19 +25,14 @@ lazy_static = "1.4.0"
 pem = "3"
 policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.21.0" }
 prettytable-rs = "^0.10"
-pulldown-cmark = { version = "0.12", default-features = false }
-pulldown-cmark-mdcat = { version = "2.7", default-features = false }
 regex = "1"
 rustls-pki-types = { version = "1", features = ["alloc"] }
 semver = { version = "1.0.22", features = ["serde"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 serde_yaml = "0.9.34"
-syntect = { version = "5.2", default-features = false, features = [
-  "default-syntaxes",
-  "parsing",
-] }
 tar = "0.4.40"
+termimad = "0.31.2"
 thiserror = "2.0"
 time = "0.3.36"
 tiny-bench = "0.4"
@@ -53,7 +50,6 @@ wasmparser = "0.227"
 reqwest = { version = "0", default-features = false, features = [
   "rustls-tls-native-roots",
 ] }
-indicatif = "0.17"
 
 [dev-dependencies]
 assert_cmd     = "2.0.14"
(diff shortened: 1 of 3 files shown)


jvanz (approved):

Cargo.toml
```diff
@@ -50,10 +44,12 @@ wasmparser = "0.227"
 # feature across all the transitive dependencies of kwctl
 # This is required to have kwctl use the system certificates instead of the
 # ones bundled inside of rustls
+indicatif = "0.17"
+is-terminal = "0.4.16"
 reqwest = { version = "0", default-features = false, features = [
   "rustls-tls-native-roots",
 ] }
-indicatif = "0.17"
+termimad = "0.31.2"
```
flavio, line 52: Can you move the new dependecies above, like in the block that ends around line 47?

The current code messes up the comment that explains why we have this "strange" `reqwest` requirement around.

src/inspect.rs
```diff
@@ -124,7 +120,8 @@ impl MetadataPrinter {
                     self.print_metadata_context_aware_resources(metadata, no_color)?;
                     println!();
                 }
-                self.print_metadata_usage(metadata, no_color)
+                self.print_metadata_usage(metadata, no_color);
+                Ok(())
```
flavio, line 124: You can remove this line and remove the one on 112. Then change the signature of this method to not be just void. We don't need to have `Result<()>` anymore, since the new code is infallible

src/inspect.rs
```diff
@@ -197,7 +193,8 @@ impl MetadataPrinter {
         table.printstd();
 
         let text = format!("```yaml\n{rules_yaml}```");
-        self.render_markdown(&text, no_color)
+        self.render_markdown(&text, no_color);
+        Ok(())
```
flavio, line 197: Same here, this method is now infallible

src/inspect.rs
```diff
@@ -219,13 +216,13 @@ impl MetadataPrinter {
         );
 
         let text = format!("```yaml\n{resources_yaml}```");
-        self.render_markdown(&text, no_color)?;
+        self.render_markdown(&text, no_color);
         println!("To avoid abuses, review carefully what the policy requires access to.");
 
         Ok(())
```
flavio, line 222: Same here, now this is infallible

flavio (changes requested):
I like the change, there are just some minor things to address

Cargo.toml
viccuad: Argh, with the rebase to main and the `indicatif` addition, I messed up with taplo.
Amended the commit.

src/inspect.rs
viccuad: for some of those `print_metadata` we still serialize with serde, or check the protocol_version and error, so I think we can't do that.

src/inspect.rs
viccuad: We serialize the rules with serde though.

flavio (approved):
Sorry about the wrong requests with `Result<()>`, it's a bit hard to get it straight by expanding the GH PR diff :sweat:

PR #1139, opened 2025-03-18
feat!: change how artifacthub scaffold works

flavio:
Scaffolding the artifacthub-pkg file relies on annotations put inside of the `metadata.yml` file.

Because of that, some flags of the `scaffold artifacthub` command have been either removed or are no longer used.

The flag that was removed has been introduced with the 1.22.1 patch release, and is probably used only by the Kubewarden team. It was required to publish the rego policies from the monorepo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -307,14 +307,14 @@ dependencies = [
 ]
 
 [[package]]
-name = "backoff"
-version = "0.4.0"
+name = "backon"
+version = "1.4.0"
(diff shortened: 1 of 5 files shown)


viccuad (approved):
Added refreshed cli-docs.md
Test coverage has a minimum decrease, but the change is already tested in policy-evaluator, and there's no test for ``kwctl scaffold artifacthub` because it's not much useful.

Approving.

PR #1157, opened 2025-03-29
feat:Added a validate_policy_title function that validates policy titles according to Kubernetes RFC 1123

### LINKED ISSUE #699: Feature: kwctl should check for validity of `metadata.name`
## Problem

Setting a wrong `annotations."io.kubewarden.policy.title"` in `metadata.yaml` may result in a wrong `metadata.name` of the CRD after scaffolding. For example a name `my policy` is invalid as it contains spaces, while `my-policy` is valid. Valid k8s `metadata.name` must follow:

> a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

This is not necessarily true for raw policies as they don't need to be deployed via a CRD in the K8s api. But maybe is good to enforce it anyways.

Notice that a user can manually edit the metadata.yaml after scaffolding.

## Acceptance criteria

- Running E2E tests with kwctl (for example `kwctl run`, `kwctl scaffold`) should error and inform the user of incorrect metadata.

maanugh:
## Description

Modified the manifest function to validate the policy title before creating the scaffold data.
Fix #699 

## Test

### DIFF
--- a/src/scaffold/manifest.rs
+++ b/src/scaffold/manifest.rs
@@ -116,6 +116,19 @@ fn build_objmetadata(data: ScaffoldPolicyData) -> ObjectMeta {
     }
 }
 
+fn validate_policy_title(title: &str) -> Result<()> {
+    // Kubernetes RFC 1123 subdomain validation regex
+    let valid_name_regex = regex::Regex::new(r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$").unwrap();
+    
+    if !valid_name_regex.is_match(title) {
+        return Err(anyhow!(
+            "Invalid title '{}'. Must use lowercase alphanumeric chars, '-' or '.', start/end with alphanumeric.",
+            title
+        ));
+    }
+    Ok(())
+}
+
 pub(crate) fn manifest(
     uri_or_sha_prefix: &str,
     resource_type: ManifestType,
@@ -135,9 +148,16 @@ pub(crate) fn manifest(
 
     let settings_yml: serde_yaml::Mapping = serde_yaml::from_str(settings.unwrap_or("{}"))?;
 
+    let policy_title = get_policy_title_from_cli_or_metadata(policy_title, &metadata);
+    
+    // Validate policy title if present
+    if let Some(title) = &policy_title {
+        validate_policy_title(title)?;
+    }
+    
     let scaffold_data = ScaffoldPolicyData {
         uri,
-        policy_title: get_policy_title_from_cli_or_metadata(policy_title, &metadata),
+        policy_title,
         metadata,
         settings: settings_yml,
     };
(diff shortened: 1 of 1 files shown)


viccuad (approved):

src/scaffold/manifest.rs
```diff
@@ -116,6 +116,19 @@ fn build_objmetadata(data: ScaffoldPolicyData) -> ObjectMeta {
     }
 }
 
+fn validate_policy_title(title: &str) -> Result<()> {
+    // Kubernetes RFC 1123 subdomain validation regex
+    let valid_name_regex = regex::Regex::new(r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$").unwrap();
+    
+    if !valid_name_regex.is_match(title) {
+        return Err(anyhow!(
+            "Invalid title '{}'. Must use lowercase alphanumeric chars, '-' or '.', start/end with alphanumeric.",
```
viccuad, line 125: I would mention RFC 1123 as part of the error.

maanugh: fixed it to 

`"Invalid title '{}'. Must conform to RFC 1123: use lowercase alphanumeric chars, '-' or '.', and start/end with an alphanumeric character.",`

src/scaffold/manifest.rs
```diff
@@ -116,6 +116,21 @@ fn build_objmetadata(data: ScaffoldPolicyData) -> ObjectMeta {
     }
 }
 
+fn validate_policy_title(title: &str) -> Result<()> {
+    // Kubernetes RFC 1123 subdomain validation regex
+    let valid_name_regex =
+        regex::Regex::new(r"^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$")
+            .unwrap();
```
flavio, line 123: The validation itself is more complex than that. This can be seen also by looking at how Kubernetes implements that ([see here](https://github.com/kubernetes/kubernetes/blob/beef784c51f466e1a021a30c2d99132a94c233b1/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L188)).

I would rewrite the code to make use of [hostname_validator](https://docs.rs/hostname-validator/latest/hostname_validator/index.html), which seems to be well done

flavio (changes requested):

src/scaffold/manifest.rs
maanugh: The key difference is that Kubernetes enforces [lowercase-only hostnames](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) (e.g., my-policy) while standard RFC 1123 validation (like [hostname_validator](https://github.com/pop-os/hostname-validator)) also allows uppercase letters (e.g., My-Policy).

we could later explicitly add Kubernetes’ extra rules if we use hostname_validator.

flavio: > The key difference is that Kubernetes enforces [lowercase-only hostnames](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) (e.g., my-policy) while standard RFC 1123 validation (like [hostname_validator](https://github.com/pop-os/hostname-validator)) also allows uppercase letters (e.g., My-Policy).

Thanks for pointing that out, I didn't realize that.

> we could later explicitly add Kubernetes’ extra rules if we use hostname_validator.

I like the idea

maanugh: please checkout the new changes

flavio (approved):
thanks!

Issue #1162, opened 2025-04-02
kwctl: building fails on windows


flavio:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Building on windows is failing, apparently due to some issue with aws-lc-sys

### Steps To Reproduce

These are the relevant bits from [this CI run](https://github.com/kubewarden/kwctl/actions/runs/14213455277/job/39824804387):

```console
Compiling icu_collections v1.5.0
warning: aws-lc-sys@0.25.1: Building with: CMake
warning: aws-lc-sys@0.25.1: Symbol Prefix: Some("aws_lc_0_25_1")
warning: aws-lc-sys@0.25.1: CMAKE environment variable set: cmake
[...]

PR #1166, opened 2025-04-08
fix(ci): build on windows

flavio:
This PR introduces two changes.

## Fix windows builds

Address the build failure currently happening on windows (fixes https://github.com/kubewarden/kwctl/issues/1162).

The GitHub runners based on Windows are now using latest version of CMake (v4).

The aws-lc-sys crate builds itself using cmake, but contains a directive that has been removed from v4.
To fix the build issue, an environment variable has to be set.

## Detect OS-related issues earlier

Run `cargo check` against windows, linux and macos. This allows us to spot issues earlier, during a PR.

Otherwise the OS-related failures are going to be visible only after PRs are merged and the `build` workflow is triggered.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -211,7 +211,14 @@ jobs:
 
       - name: enable git long paths on Windows
         if: matrix.os == 'windows-latest'
-        run: git config --global core.longpaths true
+        run: |
+          echo 'CMAKE_POLICY_VERSION_MINIMUM="3.5"' >> $GITHUB_ENV
+
+      # aws-lc-sys CMakefile contains a directive that has been removed from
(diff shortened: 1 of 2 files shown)


fabriziosestito (approved):

fabriziosestito:
@flavio DCO is failing

flavio:
Do not merge, unfortunately I've to try that doing iterations. Also, I'm currently on windows, without git configured properly. I'll amend/squash all the commits once ready from my usual dev env

flavio:
Everything is ready for review. I think the branch protection rules are now confused by the fact `cargo check` is a multi-os check now.

I can fix that once the PR is merged

viccuad (approved):
LGTM

PR #1169, opened 2025-04-10
feat: Improve policy version handling and verification

### LINKED ISSUE #537: Some `kwctl` subcommands need `registry://` while others don't
Installing policies from artifacthub.io I realized the following:

```console
$ kwctl pull ghcr.io/kubewarden/policies/safe-labels:v0.1.13 # as printed in artifacthub's install button, note no registry://
$ $ kwctl scaffold manifest -t ClusterAdmissionPolicy ghcr.io/kubewarden/policies/safe-labels:v0.1.13
Error: relative URL without a base
```

This is true for other subcommands too:
```
$ kwctl rm ghcr.io/kubewarden/policies/safe-labels:v0.1.13
Error: relative URL without a base
$ kwctl rm registry://ghcr.io/kubewarden/policies/safe-labels:v0.1.13
```

maanugh:
## Description

Fix #537 

- Add version verification before policy removal
- Improve error messages for better user experience

Before -> 

```curl
Error: could not delete policy registry://ghcr.io/kubewarden/policies/disallow-service-loadbalancer: No such file or directory (os error 2)

Or

Error: relative URL without a base
```

After -> 

```

```

```
Error: Cannot find policy with uri: Policy registry://ghcr.io/kubewarden/policies/disallow-service-loadbalancer not found. 
Available versions in local store: latest, v0.1.8
```

## Test

### DIFF
--- a/src/rm.rs
+++ b/src/rm.rs
@@ -2,10 +2,17 @@ use anyhow::{anyhow, Result};
 use policy_evaluator::policy_fetcher::store::{PolicyPath, Store};
 use std::path::PathBuf;
 
+use crate::utils::LookupError;
+
 pub(crate) fn rm(uri_or_sha_prefix: &str) -> Result<()> {
-    let uri = crate::utils::map_path_to_uri(uri_or_sha_prefix)?;
+    let uri = crate::utils::get_uri(&uri_or_sha_prefix.to_string())?;
 
     let store = Store::default();
+
+    if store.get_policy_by_uri(&uri)?.is_none() {
+        return Err(anyhow!(LookupError::PolicyMissing(uri)));
+    }
+
     let policy_path = store.policy_full_path(&uri, PolicyPath::PrefixAndFilename)?;
     std::fs::remove_file(&policy_path)
         .map_err(|err| anyhow!("could not delete policy {}: {}", uri, err))?;


src/rm.rs
```diff
@@ -39,5 +41,7 @@ pub(crate) fn rm(uri_or_sha_prefix: &str) -> Result<()> {
             });
     }
 
+    println!("Successfully removed policy: {}", uri);
+
```
jvanz, line 45: Remove this message. Let's keep it as it's now in case of successful run.

src/utils.rs
```diff
@@ -45,12 +43,45 @@ pub(crate) fn map_path_to_uri(uri_or_sha_prefix: &str) -> std::result::Result<St
...
+            uri_or_sha_prefix.clone()
+        } else {
+            format!("registry://{}", uri_or_sha_prefix)
+        };
+
+        if let Some(last_part) = uri.split('/').last() {
+            if !last_part.contains(':') {
+                return Ok(format!("{}:latest", uri));
+            }
+        }
```
jvanz, line 80: I would not add the `latest` tag. kwctl can have multiple policy versions at the same time (e.g. `latest` and `v0.2.5`). If you add the latest tag automatically, this can delete the wrong policy version by accident. Let's ask the user to define the policy he/she wants to delete.

jvanz (changes requested):

src/rm.rs
maanugh: We should notify the user when a policy is removed, as there's no log—it's silently deleted.

src/utils.rs
maanugh: fixed

src/utils.rs
```diff
@@ -37,14 +37,48 @@ pub(crate) fn map_path_to_uri(uri_or_sha_prefix: &str) -> std::result::Result<St
...
-        if let Some(policy) = store.get_policy_by_sha_prefix(uri_or_sha_prefix)? {
+        if let Some(policy) = store.get_policy_by_uri(uri_or_sha_prefix)? {
             Ok(policy.uri.clone())
         } else {
             Err(LookupError::PolicyMissing(uri_or_sha_prefix.to_string()))
         }
     }
 }
 
+pub(crate) fn verify_policy_version(uri: &str) -> std::result::Result<(), LookupError> {
```
viccuad, line 48: Could you please expand on why we need this new function?

I'm not sure the nested ifs to check for tag or digest cover all cases.
Moreover, we shouldn't need to do yet another policies.iter().filter() for performance's sake, we already iterate in in `rm()`.

In addition, for merging, we need to have the tests and linter pass.

src/rm.rs
jvanz: This a common patner in Linux. On success, just return.

src/utils.rs
maanugh: Improved the error message shown when attempting to delete a non-existent policy. Previously, users would see unclear errors like "No such file or directory" or "relative URL without a base." Now, the message clearly states that the policy was not found and lists the available versions in the local store, making it easier for users to understand and resolve the issue.

src/utils.rs
```diff
@@ -37,14 +37,48 @@ pub(crate) fn map_path_to_uri(uri_or_sha_prefix: &str) -> std::result::Result<St
...
+        let base_uri = if uri.starts_with("registry://") {
+            if let Some((base, tag_or_digest)) = uri.rsplit_once(':') {
+                if !tag_or_digest.contains('/') && !tag_or_digest.contains('@') {
+                    base
+                } else {
+                    uri
+                }
+            } else {
+                uri
+            }
```
flavio, line 61: Parsing OCI references has many subtleties, I would use a [library](https://docs.rs/oci-spec/latest/oci_spec/distribution/struct.Reference.html) for that.

However I think we can drop this code, see my next comment

flavio (changes requested):
I would remove all the code that deals with finding the name of the policy that the user might have wanted.

To clarify, I'm totally happy having kwctl show only this output:

```console
Error: Cannot find policy with uri: Policy registry://ghcr.io/kubewarden/policies/disallow-service-loadbalancer. 
```

Notice I removed the closing message ` not found.`,  I find it redundant

The rationale behind my request is to keep the code as simple as possible. I don't think this a must have feature

maanugh:
Ok, I will remove it and keep it simple

viccuad:
@maanugh can you check the e2e tests failures from the CI and fix as needed? They check for backwards compatibility, and need to be passing for us to merge the PR.

jvanz (changes requested):
Thanks for the changes! I believe we are close to merge this PR. But first we need to fix the failing CI. 

Take a look in the failing CI. At the first glance, it seems that the error message changed after your change. Please, make the tests happy again. :-)

tests/e2e.rs
```diff
@@ -276,7 +276,7 @@ fn test_digest(#[case] uri: &str, #[case] expected_sha: &str) {
     is_empty()
 )]
 #[case::sha_prefix("0169", true, is_empty())]
-#[case::non_existing("non-existing", false, contains("Cannot find policy"))]
+#[case::non_existing("non-existing", false, contains("No such file or directory"))]
```
jvanz, line 279: I think we should keep the error message and handle better the error in the code. Because the old error message makes more sense and changing it is in a breaking change.

viccuad (approved):

viccuad:
@maanugh , we require commits to be cryptographically signed. Could you amend your commit so it is signed (and appears as verified in GitHub)?

src/rm.rs
```diff
@@ -3,12 +3,12 @@ use policy_evaluator::policy_fetcher::store::{PolicyPath, Store};
...
 
 pub(crate) fn rm(uri_or_sha_prefix: &str) -> Result<()> {
-    let uri = crate::utils::map_path_to_uri(uri_or_sha_prefix)?;
+    let uri = crate::utils::get_uri(&uri_or_sha_prefix.to_string())?;
 
     let store = Store::default();
     let policy_path = store.policy_full_path(&uri, PolicyPath::PrefixAndFilename)?;
     std::fs::remove_file(&policy_path)
-        .map_err(|err| anyhow!("could not delete policy {}: {}", uri, err))?;
+        .map_err(|_| anyhow!("Failed to remove policy {}: Cannot find policy ", uri))?;
```
jvanz, line 11: If we do that, we will hide other possible errors in the std lib function used to remove the file. 

I was reviewing the code more carefully and notice something. Let's take a look in the `get_uri` function:

https://github.com/kubewarden/kwctl/blob/d49f285e421e2f8ef9d709d894f856bfb2eff413/src/utils.rs#L27-L54

Note that the `get_uri` function is hitting the `or_else` which will generate a valid policy URI even if the policy does not exists. That's why the code here is reaching the line that tries to remove the file. What you have to do is something similar of what `map_path_to_uri` does. Change the code to check if the uri returned by the `get_uri` actually exists in the store _before_ calling the `remove_file` function.  You may want to use the `get_policy_by_uri` method for that:

https://github.com/kubewarden/policy-fetcher/blob/34cdd3b9136fcbd97fbaf51ec94117ff6de9a976/src/store/mod.rs#L170

Sorry for this ping-pong and thanks for your patience! :heart:

jvanz (approved):
LGTM. Thanks for your patience! :)

jvanz:
As we are in rc phase, we cannot merge this until the `v1.24` final release. Moving this to block.

flavio (approved):
thanks for having implemented this change

PR #1194, opened 2025-05-05
fix(deps): remove deprecated actions-rs/cargo Github action.

jvanz:
## Description

Removes the deprecated actions-rs/cargo Github action used to run cargo commands.

Related to https://github.com/kubewarden/kubewarden-controller/issues/1092

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,11 +34,22 @@ jobs:
           target: ${{matrix.targetarch}}-unknown-linux-musl
           override: true
 
-      - uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # v1.0.3
-        with:
-          use-cross: true
-          command: build
-          args: --release --target ${{matrix.targetarch}}-unknown-linux-musl
+      - name: Install cross-rs
+        run: |
+          set -e
+
+          echo "$CROSS_CHECKSUM  cross-x86_64-unknown-linux-musl.tar.gz" > checksum
+          curl -L -O https://github.com/cross-rs/cross/releases/download/$CROSS_VERSION/cross-x86_64-unknown-linux-musl.tar.gz
+          sha512sum -c checksum
+          tar -xvf cross-x86_64-unknown-linux-musl.tar.gz
+        env:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -34,11 +34,20 @@ jobs:
...
-          args: --release --target ${{matrix.targetarch}}-unknown-linux-musl
+      - name: Install cross-rs
+        shell: bash
+        run: |
+          set -e
+
+          echo "70b31b207e981aa31925a7519a0ad125c5d97b84afe0e8e81b0664df5c3a7978558d83f9fcd0c36dc2176fc2a4d0caed67f8cf9fd689f9935f84449cd4922ceb  cross-x86_64-unknown-linux-musl.tar.gz" > checksum
+          curl -L -O https://github.com/cross-rs/cross/releases/download/v0.2.5/cross-x86_64-unknown-linux-musl.tar.gz
+          sha512sum -c checksum
+          tar -xvf cross-x86_64-unknown-linux-musl.tar.gz
```
flavio, line 45: This is going to be a pain to maintain.

I would prefer to use `cargo-bininstall` as recommended [here](https://github.com/cross-rs/cross/?tab=readme-ov-file#installation).

If we really want to stick with the GH release, then I would create a KW action like the one we have to install kwctl.

flavio (changes requested):

.github/workflows/build.yml
jvanz: As the cargo-bininstall github action is not in our "safe" list to be used. If we decided to use cargo-bininstall we will download the binary manually or compile it from source code (`cargo install cargo-binstall`). In the first option we will still download a binary manually to later download the cross binary as well. In the second, it will take more time from the CI. That's why I've decided to download the cross binary manually here. It does not have [updates](https://github.com/cross-rs/cross/releases) since 2023. Thus, I do not expect so many updates in a recent future.

Besides that, I've decided to not create a "cargo github action" because this is the only place where we use cross-rs.  Therefore, we save some dependencies PRs bumping the github action once we bump the github actions version. But if you think this is still necessary, I can create it, no problems.  ;)

jvanz: The @kubewarden/kubewarden-developers team discussed this during a call and decided to not creating a github action to install cross and keep the installation step. But we will change to define the checksum and the version in environment variables to facilitate later on a possible automatic version bump.

viccuad (approved):
LGTM!

flavio (approved):

Issue #1299, opened 2025-08-04
`kwctl scaffold admission-request --object test.yaml` overwrites the file contents if no resource catalog cache is present


viccuad:
Running the scaffold command `kwctl scaffold admission-request --object test.yaml` correctly prints the output,
but additionally overwrites the `test.yaml` file with an empty cache contents for the resource catalog, when the cache is not present.

This is normally not triggered, as usually the cache creation succeeds. This may happen for users that don't have a local cluster running, or users that are hitting https://github.com/kubewarden/kwctl/issues/1298 though.

### Workaround

Recreate the cache by running the command with `kwctl <= 1.21` at least once. See https://github.com/kubewarden/kwctl/issues/1298.

### Reproducer

```console
$ rm -rf ~/.cache/kubewarden
$ cat test.yaml
$ kwctl-1.27 scaffold admission-request --operation CREATE --object test.yaml
$ cat test.yaml
```

This doesn't happen if there's an existing `~/.cache/kubewarden/kwctl/resource_catalog.json` (one can create it with kwctl <= 1.21, see https://github.com/kubewarden/kwctl/issues/1298).

PR #1300, opened 2025-08-04
fix: Use correct path for resource-catalog.json cache file

viccuad:
## Description

Fix https://github.com/kubewarden/kwctl/issues/1299

## Test

No need for new tests. Would be to add a test here, but we are already checking for the error, no need to check that we aren't overwriting the incorrect file:
https://github.com/kubewarden/kwctl/blob/main/src/scaffold/admission_request.rs#L771

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3151,6 +3151,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "sha2",
  "tar",
  "tempfile",
  "termimad",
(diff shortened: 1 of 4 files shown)


jvanz (approved):

jvanz (approved):

flavio:
I did a forced push because I forgot to add the fixture file used by the tests :facepalm:

flavio:
Warning, GHA has not been triggered by my push...

Issue #1298, opened 2025-08-04
`kwctl scaffold admission-request` errors when creating the resource catalog


viccuad:
After some investigation, it seems that starting from 1.22, kwctl fails when trying to create a kube client for connecting against a cluster

### Steps to reproduce

1. Start a local cluster (e.g: `minikube start`)
2. Call `kwctl`:
```console
$ cat test.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-allow-hostnetwork-true
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test-allow
  template:
    metadata:
      labels:
        app: test-allow
    spec:
      hostIPC: true
      containers:
        - name: nginx
          image: nginx:alpine

$ kwctl-1.22.0 scaffold admission-request --operation CREATE --object test.yaml
2025-08-04T10:12:17.871203Z  INFO kwctl::scaffold::admission_request: Resource catalog not found, building it

thread 'main' panicked at /cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustls-0.23.21/src/crypto/mod.rs:249:14:
no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: rustls::client::client_conn::ClientConfig::builder_with_protocol_versions
   4: rustls::client::client_conn::ClientConfig::builder
   5: <kube_client::config::Config as kube_client::client::config_ext::ConfigExt>::rustls_client_config
   6: <kube_client::config::Config as kube_client::client::config_ext::ConfigExt>::rustls_https_connector_with_connector
   7: <kube_client::client::builder::ClientBuilder<tower::util::boxed::sync::BoxService<http::request::Request<kube_client::client::body::Body>,http::response::Response<alloc::boxed::Box<dyn http_body::Body+Data = bytes::bytes::Bytes+Error = alloc::boxed::Box<dyn core::error::Error+core::marker::Send+core::marker::Sync>+core::marker::Send+core::marker::Unpin>>,alloc::boxed::Box<dyn core::error::Error+core::marker::Send+core::marker::Sync>>> as core::convert::TryFrom<kube_client::config::Config>>::try_from
   8: <kube_client::client::Client as core::convert::TryFrom<kube_client::config::Config>>::try_from
   9: kwctl::scaffold::admission_request::build_kube_client::{{closure}}
[...]

PR #1301, opened 2025-08-04
fix: Set default rustls provider

viccuad:
## Description

Fix https://github.com/kubewarden/kwctl/issues/1298

This fixes error on creating the resource catalog for the first time.

The rustls dependency is configured exactly as we do in policy-fetcher.

## Test

Added new e2e test for `kwctl scaffold admission-request`. Doubles as e2e test for https://github.com/kubewarden/kwctl/pull/1300.


Cargo.toml
```diff
@@ -51,6 +51,10 @@ hostname-validator = "1.1.1"
 reqwest = { version = "0", default-features = false, features = [
   "rustls-tls-native-roots",
 ] }
+rustls = { version = "0.23", default-features = false, features = [
```
viccuad, line 54: Maybe it's better to re-export on policy-fetcher and consume it that way.

jvanz: Fine by me. One less dependency to keep on sync among our code.

tests/e2e.rs
```diff
@@ -923,6 +923,30 @@ fn test_artifacthub_scaffold_fail_when_metadata_not_provided_nor_found() {
...
+    let mut cmd = setup_command(tempdir.path());
+    cmd.arg("scaffold")
+        .arg("admission-request")
+        .arg("--operation")
+        .arg("CREATE")
+        .arg("--object")
+        .arg(&deployment_yaml);
+    cmd.assert().success();
+
+    // Fail if deployment.yaml test data file was changed
```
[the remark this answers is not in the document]
jvanz: Let's wait to rebase it before merge it then.

Cargo.toml
```diff
@@ -51,6 +51,7 @@ hostname-validator = "1.1.1"
 reqwest = { version = "0", default-features = false, features = [
   "rustls-tls-native-roots",
 ] }
+rustls = { version = "0.23", default-features = false }
```
flavio, line 54: Have you tried to relax that even more, like `version = "0"`. This should avoid any complication with the next minor release of rustls

PR #1411, opened 2025-10-27
    fix(ci): sign blobs with cosign v3 signature bundle.

jvanz:
## Description

Updates the CI to signed the blobs using only cosign v3 signature bundle format.


viccuad (approved):
LGTM

